PR #14, opened 2015-01-12
Update event urls to remove catch-all url

### LINKED ISSUE #13: catch-all url for event
Hi

In core/urls.py, there is this line:
url(r'^(?P<city>[\w\d/]+)$', 'core.views.event', name='event'),

This basically says "match djangogirls.org/ plus any combination of letters and go to the event view".  Then in core/views.py, there is a try/except clause on line 50 that says if the event page doesn't exist, redirect to /events.  When developing the jobs site, we've found that the url for /jobs/ redirects to /events when the trailing slash is omitted - the event url plus the redirect is a catchall url, so we can never get a 404 (any nonsense url e.g. djangogirls.org/sdfshdjkfkasejaoefjasdlk redirects to djangogirls.org/events).  

I can see that a workaround redirect has already been done for /admin in djangogirls/urls.py, but if the site is being expanded with more apps/sections, this will mean that it will need to be done for every url.  Is this how you'd prefer it to be done, or would it be better to add a prefix for event (/event/berlin etc - which admittedly will need to deal with the event pages already set up)

Thanks
Becky

rebkwok:
Resolves #13 
Updated the url for event so that an unrecognised url now raises a 404 which shows the all events link (rather than just redirecting to the events page).  Also added different pages for unpublished events depending on whether past or future, which include links to the all events page and an email link for the organiser of the requested event.

### DIFF
--- a/core/fixtures/core_views_testdata.json
+++ b/core/fixtures/core_views_testdata.json
@@ -101,6 +101,22 @@
 },
 {
     "fields": {
+        "city": "Hidden_no_date",
+        "name": "Django Girls Hidden",
+        "country": "Event",
+        "latlng": null,
+        "team": [
+            1
+        ],
+        "is_on_homepage": false,
+        "main_organizer": 1,
+        "date": ""
+    },
+    "model": "core.event",
+    "pk": 4
+},
+{
+    "fields": {
         "custom_css": null,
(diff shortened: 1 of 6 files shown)


olasitarska:
Fantastic work! Thank you so much @rebkwok! :sparkles: :heart:

Issue #19, opened 2015-03-11
Displaying detail of application and scoring for organizers


olasitarska:
> This issue is a part of [Project: application process](https://github.com/DjangoGirls/djangogirls/wiki/Project:-Application-process) & the work is happening on [applications branch](https://github.com/DjangoGirls/djangogirls/tree/applications)

Once we have a list of all applications for organizers (#18), then we need to create an application detail view. This also needs to be restricted to organizers of the city only. 

On this page, you should see:
- all the questions & answers for the application
- at the end, possibility to score application in the scale of 1-5 and skip scoring for this application. Organizers should also have a way to add a comment to their score.

Buttons:
- save
- save and go to next application

Scores of other organizers should not be visible until you score the application yourself.

This also requires adding a model with scores.

phalt:
I'm going to (hopefully) complete this view today, along with the ability to score the application. It won't look pretty but the backend functionality shouldn't take me more than a few hours.

asendecka:
I keep my fingers crossed! :+1:

Issue #18, opened 2015-03-11
Displaying list of applications for organizers with sorting


olasitarska:
> This issue is a part of [Project: application process](https://github.com/DjangoGirls/djangogirls/wiki/Project:-Application-process) & the work is happening on [applications branch](https://github.com/DjangoGirls/djangogirls/tree/applications)

The goal of this is to have a list of all `Application` objects for given `Form` listed somewhere. Applications should be listed anonymously, with just an ID and date of creation. 
Url: http://djangogirls.org/city/applications/
Access should be restricted to organizers of this event only. 

This is a first part of the scoring / triaging system.

phalt:
Gonna give this a go today. I'm not a css whiz, so I'll just start some work on providing lists of applications for events, filtered on events and only viewable by organisers of that event.

andyosuna:
If this is okay I can take over and help with the styling!

phalt:
@andyosuna Sure, I think @olasitarska merged the progress I did into the `applications` branch

olasitarska:
Yes @andyosuna, go ahead! :)

andyosuna:
I believe the applications shouldn't link to a detail page, like they do now. Only styled the applications list for now :smile_cat:

phalt:
@andyosuna The link goes to the detail view described [in this issue](https://github.com/DjangoGirls/djangogirls/issues/19). We can change it to a button instead, however.

phalt:
@olasitarska What type of sorting do we want for this view? Sorting based on the state of the applications? newest/oldest/accepted/rejected?

olasitarska:
I think we only need sorting by score for now -- I added it now :)

PR #28, opened 2015-03-14
View list and detail applications for organisers.

phalt:
Redacted

### DIFF
--- a/applications/utils.py
+++ b/applications/utils.py
@@ -27,13 +27,11 @@ DEFAULT_QUESTIONS = [
         "title": "Which operating system do you use?",
         "question_type": "choices",
         "choices": "Mac OS X; Windows; Linux",
-        "has_option_other": True,
     },
     {
         "title": "What is your current level of experience with programming?",
         "question_type": "choices",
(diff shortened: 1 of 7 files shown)


phalt:
This Pull Request adds the following features:

- View for seeing all applications submitted to a particular page/city/event (Partial for #18)
- View for seeing the detail view of application questions and answers (Partial for #19)
- A decorator for views that allows only super user and members of the page/city/event team to view. Returns a 404 Not Found if you do not meet these criteria.

There is no sorting for applications or the ability to triage them yet. I'm absolutely terrible with front end pretty things, so I've just written the basic html tag stuff. I hope this helps!

applications/utils.py
```diff
@@ -27,13 +27,11 @@
         "title": "Which operating system do you use?",
         "question_type": "choices",
         "choices": "Mac OS X; Windows; Linux",
-        "has_option_other": True,
```
phalt: These were breaking after being removed in a previous migration :)

andyosuna: I commented those lines out and am working like that xP

olasitarska:
This looks amazing @phalt! Obviously some css definitely missing from the templates, but we can take care of that in another issue :) 

Thank you, you rock! :+1: :tulip: :koala:

PR #33, opened 2015-03-22
Actually fix pep8 in applications

phalt:
I only ran flake8 on a few files last time. Sorry :(

This time I ran it on the entire applications/ directory. There is still a single error in the applications/tests.py file, but that should be fixed as soon as people start writing some tests ;)


asendecka:
Awesome! Thank you! :bicyclist: :dancer:

PR #46, opened 2015-06-04
Upgrade to latest stable security release in 1.7 release line.

robcharlwood:
Upgrade to latest security release.


aniav:
I believe we can also move to Django 1.8 if someone is interested :zap:

PR #47, opened 2015-06-04
Turned off template debug in prodution settings

robcharlwood:
Tighten up all the security things! :)


olasitarska:
:rocket: :heart_eyes: thanks!

PR #54, opened 2015-06-04
Upgrading to Django 1.8

MounirMesselmeni:


olasitarska:
@MounirMesselmeni Could you also update the requirements file to bump it to 1.8? This is how Heroku install requirements too :)

olasitarska:
Thank you! :sparkles:

PR #56, opened 2015-06-05
Applications

parodia:
Organizer can customize the message which is sent as the confirmation of application.
Also the event_mail is added to model Event (can be changed only by superadmin) - which is used to send the confirmation mails.


olasitarska:
Thanks Justyna! :heart:

PR #62, opened 2015-06-21
Adding a Makefile

rebecabordini:
Adding a makefile to the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ bin
 lib
 src
 include
+.idea/
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,29 @@
+all: help
+
+help:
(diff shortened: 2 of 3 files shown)


aniav:
Hmm, I am not really sure if this addition is desired. There is a couple of reasons:
* the commands aren't a lot shorter than the original ones
* it is not common to create makefiles for Django projects - so it will make harder for newcomers to understand what's going on
* there are no common commands like `make install` in the makefile

If I am missing something about this change being a good addition - please correct me :sparkles: :+1:

olasitarska:
I agree with Ania!

We also run the python manage.py new_event very often on Heroku (with `heroku run python manage.py new_event`), so shortcut will only work locally, and it'll make it more confusing, I believe.

PR #63, opened 2015-06-23
Fix the order of the questions in the form

### LINKED ISSUE #60: Order of the questions in the form can get mixed up
Haven't debug it yet, but just adding so I won't forget

hawkowl:
Python dicts aren't sorted, but an OrderedDict is, so we should use that instead.

Fixes #60.


olasitarska:
:+1: :sparkles: thanks!

PR #67, opened 2015-06-25
Convert __unicode__ to __str__ in some models

helenst:
converting some __unicode__ things to __str__ for easier navigation of the admin in py3


olasitarska:
Thanks! :) :koala:

PR #76, opened 2015-07-05
Correcting applications Application model test (test_average_score)

MarysiaLowas:
To get an average score, the test draw a random digit between 0 and 5.
But the average_score property in the Application model takes only scores with are not 0 (if s.score).

This makes the test unstable.
If the test draws 0 and 5, the average with be 2.5 for the test but 5 for the model (it will ignore 0)
and the test will fail.

I've changed the test to draw a random digit between 1 and 5.


olasitarska:
Thanks @MarysiaLowas! :)

PR #78, opened 2015-07-07
Fix menu entries in the applications pages

### LINKED ISSUE #80: Links in Applications menu don't work
The links to "Applications" and "Messaging" from the /applications page lead to ```djangogirls.org/event_city//event_city/applications``` and ```djangogirls.org/event_city//event_city/communication```, respectively. See screenshot. The result is a 500 error. 

I think the issue is in /templates/event/menu.html, or in how the data gets passed in and then rendered in line 19, and in applications/views.py, lines 72-79.  

Not a huge deal, and I know a lot of us are traveling this weekend. :-) Just wanted to let someone know! 

screen shot 2015-07-08 at 9 39 59 pm

helenst:
Menu entries have been broken in the organiser-visible application
pages, the problem is that the city part of the URL goes in twice. This
strips it out before it gets sent off to the template :tulip:

fixes #80

### DIFF
--- a/applications/tests/tests_utils.py
+++ b/applications/tests/tests_utils.py
@@ -1,0 +1,9 @@
+from django.test import TestCase
+from applications.utils import get_organiser_menu
+
+
+class MenuTest(TestCase):
+    def test_menu_entries(self):
+        menu = get_organiser_menu('london')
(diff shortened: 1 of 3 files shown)


helenst:
closing this off as @uranusjr is handling this in #81

PR #81, opened 2015-07-09
Make menus in event-related pages capable of showing absolute path and URI

uranusjr:
<http://djangogirls.org/taipei/> has links to external pages in the menu (the third one from the left), and simply joining path components doesn’t work.

I use a custom template tag to detect absolute URLs (using `urllib.parse.urlparse`) and choose the right behaviour accordingly.

### DIFF
--- a/applications/tests/test_utils.py
+++ b/applications/tests/test_utils.py
@@ -1,0 +1,11 @@
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+
+from applications.utils import get_organiser_menu
+
+
+class MenuTest(TestCase):
+    def test_organiser_menu_entries(self):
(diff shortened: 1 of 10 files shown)


uranusjr:
Wow yes, I didn’t read #78 and #80, but these two issues should probably be solved together. With this approach, URLs of menu items will always be either absolute, external, or relative to the event page.

With that said, I find that my change actually breaks the application list page. Shame on me for not running tests. I will fix that first.

helenst:
Awesome - thanks for doing this. I just tried it out and it seems great! :peach:

olasitarska:
Thaaaaaaanks! :rose:

williln:
Thanks @uranusjr! That's awesome!

PR #84, opened 2015-07-11
Automation magic for new_event command 🎩🐇

olasitarska:
So basically I 
- refactored the whole command a bit. I'm still not super happy with the state of it, but ¯\_(ツ)_/¯
- added Slacker package
- added functionality that invites organizers to Slack automatically
- added fancy colors!
- added a bot that messages Slack #general channel when new event is added

### DIFF
--- a/core/default_eventpage_content.py
+++ b/core/default_eventpage_content.py
@@ -36,40 +36,40 @@ def get_default_eventpage_data():
             'name': 'about',
             'is_public': True,
             'background': get_random_photo('about'),
-            'content': '<div style="text-align: center;"><h1>Free programming workshop for women</h1><h2>Build your first website at EuroPython 2014 in Berlin!</h2><a class="btn" href="#value">Learn more »</a></div>'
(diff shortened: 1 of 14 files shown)


core/default_eventpage_content.py
```diff
@@ -0,0 +1,84 @@
...
+    return None
+
+
+def get_default_eventpage_data():
+    return [
+        {
+            'name': 'about',
+            'is_public': True,
+            'background': get_random_photo('about'),
+            'content': '<div style="text-align: center;"><h1>Free programming workshop for women</h1><h2>Build your first website at EuroPython 2014 in Berlin!</h2><a class="btn" href="#value">Learn more »</a></div>'
```
aniav: I believe putting in a template and loading it here would be much more readable for further changes. Maybe a #TODO if you don't have time for that now? :)

olasitarska: Totally make sense. I'll do that, I'm on sprints today! :)

PR #85, opened 2015-07-12
Change "state" to "status"

williln:
We're using "status" later in the page, so for consistency, I'd like to change "state" to "status." Also, "Choose state" sort of made me think that I could see the applications from Texas versus Oklahoma, so "status" is a little clearer (to me, anyway) :-)


olasitarska:
Thanks Lacey! :) :panda_face:

PR #88, opened 2015-07-15
Corrected typo

williln:
There was a period after an exclamation point, and I removed it.


asendecka:
Thanks! :chestnut: :cherries:

PR #91, opened 2015-07-21
Allow event map to filter on past / upcoming event

helenst:


templates/events_map.html
```diff
@@ -10,7 +10,13 @@
    
     <script>
         L.mapbox.accessToken = 'pk.eyJ1Ijoib2xhc2l0YXJza2EiLCJhIjoiZGUzZjUyNjUxYWNkOGU1ODBkYjcxNzc4MTllNGRlMjcifQ.t0yuSZmW5FvragLmVgJXNQ';
-        var map = L.mapbox.map('map', 'olasitarska.m8nged0f', {scrollWheelZoom: false}).setView([30, 0], 2);
+        var past = L.layerGroup()
+        var future = L.layerGroup()
+        var map = L.mapbox.map('map', 'olasitarska.m8nged0f', {scrollWheelZoom: false, layers: [past, future]}).setView([30, 0], 2);
```
aniav: I know this was already in the code but shouldn't the string `olasitarska.m8nged0f` be placed in the settings? :)

helenst: Probably should be :) I'll have a look at it while I'm in this bit of code.

aniav:
Bam! :sparkles:

Issue #86, opened 2015-07-14
Validation of application scoring form


olasitarska:
![screen shot 2015-07-14 at 9 47 18 am](https://cloud.githubusercontent.com/assets/527795/8676396/d9797c2a-2a0e-11e5-8833-570c07c09b61.png)

williln:
@olasitarska What do you want to happen here? Have it average out to 0.5?

olasitarska:
Sorry for not making that clear, here it is:
- In the row with empty person -- this person doesn't have first_name and last_name setup. In this case, we should display email field.
- Ola Sendecka and empty person shouldn't be allowed to add a score with score None and comment None.

aniav:
This is also done, closing!

PR #94, opened 2015-07-22
Added display of email when scorer has no first/last name; hid other …

williln:
…scores when user's score isn't >0.

Addresses part of https://github.com/DjangoGirls/djangogirls/issues/86. User can still add score of None and comment of None, but you can no longer save the application with a None score and still see the other scores.

### DIFF
--- a/templates/application_detail.html
+++ b/templates/application_detail.html
@@ -39,12 +39,7 @@
                 </thead>
                 {% for s in scores %}
                 <tr>
-                    <td>{% if s.user.first_name or s.user.last_name %} 
-                            {{ s.user.first_name }} {{ s.user.last_name }}
-                        {% else %}
-                            {{ s.user.email }}
-                        {% endif %} 
-                    </td>
+                    <td>{{ s.user }}</td>
                     <td>{{ s.score }}</td>
                     <td>{{ s.comment }}</td>
                 </tr>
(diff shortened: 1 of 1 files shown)


templates/application_detail.html
```diff
@@ -39,7 +39,12 @@ <h3 class="page-header">Application #{{ application.number }} for {{ page.title
                 </thead>
                 {% for s in scores %}
                 <tr>
-                    <td>{{ s.user.first_name }} {{ s.user.last_name }}</td>
+                    <td>{% if s.user.first_name or s.user.last_name %}
```
aniav: I think this should be taken from `_str_` method on the user model :)

williln:
@aniav Changed to use just s.user :smile: Now it looks like this: 

screen shot 2015-07-22 at 11 15 00 am

aniav:
Well, it seems ok, I think. It could be improved on the model side but I think I will just merge it now :D

williln:
I'm happy to keep working on it; I want to address @olasitarska's note about users shouldn't be able to save when both score and comments are null, anyway. What do you have in mind? __str__ should show name if there is a name, and email if there isn't?

aniav:
Yeah, I think it should be checked and if both first name and last name are empty it should just siply display the e-mail, and I feel this can be changed in `_str_` as this is the behaviour we would always like to have for representing the user :)

PR #95, opened 2015-07-22
Changed User model to return email if no first/last name 

williln:
Tried to just check the result of ```get_full_name``` but that didn't work for some reason.

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -47,7 +47,10 @@ class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
         verbose_name_plural = 'Organizers'
 
     def __str__(self):
-        return '{0} ({1})'.format(self.get_full_name(), self.email)
+        if self.first_name == '' and self.last_name == '': 
+            return '{0}'.format(self.email)
+        print('Name:', self.get_full_name())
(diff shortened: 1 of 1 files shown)


core/models.py
```diff
@@ -47,7 +47,9 @@ class Meta:
         verbose_name_plural = 'Organizers'
 
     def __str__(self):
-        return '{0} ({1})'.format(self.get_full_name(), self.email)
+        if self.first_name == '' and self.last_name == '': 
+            return '{0}'.format(self.email)
+        return '{0}'.format(self.get_full_name())
```
aniav: I think that the e-mail should still be attached to the full_name if it is there. I am affraid it can be used somewhere for copy&paste reasons for e-mails or something :)

Sorry if I sound picky, I hope you will forgive me :)

aniav:
:sparkling_heart: :+1: :tulip:

williln:
No worries! Like this? This is the output: 

screen shot 2015-07-22 at 11 40 03 am

williln:
Wow you are fast @aniav!

aniav:
Bam! :D (no special skills, I just had this website open and it refreshed automatically :P)

PR #97, opened 2015-07-24
Disable score save if score = 0

williln:
@bmispelon or @aniav or @olasitarska or anyone else, this obviously isn't ready (cause it doesn't work!) but I want to address https://github.com/DjangoGirls/djangogirls/issues/86, and I'm having trouble because I don't really work with models (which is a whole discussion about doing Django on a mainframe that's neither here nor there...) 

I want to address @olasitarska's comment about not letting people add a score of None and a comment of None and then save, but I'm having trouble. Can you take a look at this code and give me some guidance? Please and thank you! :heart_decoration:


olasitarska:
Ok, so what I think we should do here:

1) From `Score` model in `applications/models.py`, in a field `score`, we should remove `null=True, blank=True`. This will make this field required.
2) In the same field, change `MinValueValidator` from 1 to 0. 0 should be a possible option that basically means: I don't want to vote on this person, and this vote shouldn't be taken into account in average score. 
3) On the same field, we should add `default=0`, so that when we make a migration, it'll know that it should change all empty fields to 0. 
4) Then we should make a migration with `python manage.py makemigrations` and `python manage.py migrate`

And I think that should generally solve all validation problems -- it would no longer be possible to add Score object with no score, because Django will take care of all validation. 

But, we're missing one more bit:

In Application model, method `average_score`, we've got something like this: 
```
scores = [s.score for s in self.scores.all() if s.score]
```
We should add a condition here to only take into account scores that are higher than 0. 

Hope that helps! <3

williln:
Thank you! It does, because I also misunderstood something-- I thought we wanted them to be able to save notes without a score or a score without notes, but we don't, which makes this a little easier :confetti_ball: I'll work on this over the weekend!

olasitarska:
Awesome, thanks! :)

williln:
@olasitarska et al., I started to work on this tonight, realized I already had (and so my last 2 commits are sort of useless), and tested it. I think this is ready to be merged! Sorry this took so long; lemme know if anything looks wonky.

applications/views.py
```diff
@@ -3,7 +3,6 @@
 from django.http import Http404, JsonResponse, HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.contrib import messages
-from django.template.defaultfilters import striptags
```
aniav: Removing this breaks the tests.

templates/application_detail.html
```diff
@@ -9,6 +9,7 @@
 <div class="container without-background">
     <div class="row">
         <div class="col-md-12 applications">
+            {{ form.name_of_field.errors }}
```
aniav: I am not sure why is this here. I am pretty sure this won't work.

aniav:
This basically works. @williln could you please fix the two small errors I found and then we will be ready to merge :)))

aniav:
Ok, I am merging it now, and I will fix the small issues as a separate commit so we have it closed :)

Issue #101, opened 2015-07-28
Applications CSV export: add button "Export to CSV"


jinpark:
I already got started on this so I'll try to have a PR for this sometime soon

olasitarska:
Ah, sorry, I duplicated this in #105 :)

jinpark:
yay!

PR #102, opened 2015-07-29
add method for downloading applicant list

jinpark:
For https://github.com/DjangoGirls/djangogirls/issues/101

if you hit `/city/applications/download`, you get a csv with applicant id, email, state, rsvp status, average score

Added tests so test coverage doesn't go down because of me :)

### DIFF
--- a/applications/tests/tests_views.py
+++ b/applications/tests/tests_views.py
@@ -1,4 +1,6 @@
 import json
+from io import StringIO
+import csv
 from datetime import timedelta
 
 from django.test import TestCase, RequestFactory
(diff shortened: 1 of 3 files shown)


olasitarska:
This looks awesome @jinpark! It got me thinking -- should we actually try to return all questions and responses too? It maybe helpful, for example for adding people to groups based on their level of knowledge or operating system.

jinpark:
Yea, that sounds good! I kept it simple because I wasn't sure what we wanted here. I can give it a shot with the responses too. I'll try to have something ready soon :)

olasitarska:
Awesome! I'll merge it now, but looking forward to another PR from you soon!

PR #109, opened 2015-07-31
add actual button to application page for downloading to csv

jinpark:
For https://github.com/DjangoGirls/djangogirls/issues/101

Not sure if we compile `.styl` files when we deploy but I added the `css` file anyway. Not sure why theres so many changes when I changed just one thing in the styl file. :P

### DIFF
--- a/templates/applications.html
+++ b/templates/applications.html
@@ -40,7 +40,7 @@
               </select>
               <input type="submit" value="Apply" />
           </p>
-          <a href="download" class="btn download">Download to csv</a>
+          <a href="{% url 'applications:applications_csv' page.url %}" class="btn download">Download to csv</a>
           <p>Showing {{ applications.count }} applications out of {{ all_applications_count }}:</p>
           <table class="table">
               <thead>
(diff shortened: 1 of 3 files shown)


templates/applications.html
```diff
@@ -40,6 +40,7 @@ <h3 class="page-header">Applications for {{ page.title }}</h3>
               </select>
               <input type="submit" value="Apply" />
           </p>
+          <a href="download" class="btn download">Download to csv</a>
```
helenst: It would be better to use the `{% url  %}` tag here for the href :strawberry:

jinpark: Ahh, good call. Let me make that change

jinpark: Made the change :)

helenst:
Looks awesome to me. I'll leave for somebody else to merge as I'm not actually sure about the CSS thing. :sparkling_heart:

olasitarska:
Yeap, that is correct -- we commit css too, because we don't have a preproccessor as part of deploy yet.

olasitarska:
Thanks @jinpark and @helenst for review! :sparkles:

Issue #106, opened 2015-07-30
Application CSV export: add all questions and answers to exported CSV


olasitarska:
At the moment, exported CSV contains only a couple of columns. We would like to have all questions and answers exported too :) 

cc @jinpark

jinpark:
whoo!

PR #110, opened 2015-07-31
add questions and answers to csv download

jinpark:
For https://github.com/DjangoGirls/djangogirls/issues/106

Adds all questions to the csv header and answers to the csv rows


applications/views.py
```diff
@@ -98,9 +99,14 @@ def applications_csv(request, city):
     response = HttpResponse(content_type='text/csv')
     response['Content-Disposition'] = u'attachment; filename="{}.csv"'.format(city)
     writer = csv.writer(response)
-    writer.writerow(["Application Number", "email", "Application State", "RSVP Status", "Average Score"])
+    csv_header = ["Application Number","Application State", "RSVP Status", "Average Score"]
+    # replace commas with semicolons so it won't confuse csv parsers
+    csv_header.extend([striptags(question).replace(',', ';') for question in Question.objects.values_list('title', flat=True)])
```
helenst: This gets you all the questions... it needs to be restricted to just that application form. (try having two events in the system and see what I mean :smile:)

I am also not sure about the commas needing to be replaced... I think `csv` takes care of this and surrounds it with quotes.

phalt: Just a small addition, are we trying to be PEP8 complaint? This a very long line :yum:

jinpark: Heh, lemme fix both these problems :)

phalt: :dancer:

jinpark: Looked into it and the csv module doesn't seem to escape the commas. And I made it so it only gets the questions from the specific page/form. AND also shortened up the line length :)

Thanks for the review!

helenst: Great - it seems to be loading the correct set of questions now :confetti_ball:

Re. the CSV - the csv module will surround the value with quotation marks so that csv readers will interpret it correctly. It shouldn't need anything else.

Are you seeing this cause problems with any particular CSV reader?

jinpark: I thought I did but then I realised I was loading the wrong file :P

Made the changes :)

applications/views.py
```diff
@@ -98,9 +99,16 @@ def applications_csv(request, city):
     response = HttpResponse(content_type='text/csv')
     response['Content-Disposition'] = u'attachment; filename="{}.csv"'.format(city)
     writer = csv.writer(response)
-    writer.writerow(["Application Number", "email", "Application State", "RSVP Status", "Average Score"])
+    csv_header = ["Application Number","Application State", "RSVP Status", "Average Score"]
+    questions = page.form_set.first().question_set.values_list('title', flat=True)
+    # replace commas with semicolons so it won't confuse csv parsers
```
helenst: can remove this comment now

applications/views.py
```diff
@@ -98,9 +99,16 @@ def applications_csv(request, city):
...
+    csv_header.extend([striptags(question) for question in questions])
+    writer.writerow(csv_header)
     for app in applications:
-        writer.writerow([app.number, app.email, app.state, app.rsvp_status, app.average_score])
+        app_info = [app.number, app.state, app.rsvp_status, app.average_score]
+        app_answers = app.answer_set.values_list('answer', flat=True)
+        app_info.extend([answer for answer in app_answers])
```
helenst: No need for the list comprehension any more

applications/views.py
```diff
@@ -98,9 +99,16 @@ def applications_csv(request, city):
     response = HttpResponse(content_type='text/csv')
     response['Content-Disposition'] = u'attachment; filename="{}.csv"'.format(city)
     writer = csv.writer(response)
-    writer.writerow(["Application Number", "email", "Application State", "RSVP Status", "Average Score"])
+    csv_header = ["Application Number","Application State", "RSVP Status", "Average Score"]
+    questions = page.form_set.first().question_set.values_list('title', flat=True)
+    # replace commas with semicolons so it won't confuse csv parsers
+    csv_header.extend([striptags(question) for question in questions])
```
helenst: You could use `map(striptags, questions)` as a shorter alternative here - probably a matter of preference, so it's just a suggestion :wink:

helenst:
This is really awesome - we are just about to need this for Coventry and it will make things sooo much easier! Just a few tiny little final tweaks and then we can merge :balloon::tiger:

applications/views.py
jinpark: done!

applications/views.py
jinpark: done!

applications/views.py
jinpark: Yea, map makes sense here :)

jinpark:
@helenst made the changes you requested, should be ready to go!

helenst:
:sparkles: :sparkling_heart: You are awesome, thank you for your work on this! :dolphin:

Issue #117, opened 2015-08-10
Fix weird ordering in /applications/ page


jinpark:
On the applications page, the ordering is wonky and I'm not really sure why

![image](https://cloud.githubusercontent.com/assets/1847812/9162321/4d49c57a-3f80-11e5-8460-33347f384d92.png)

phalt:
Could this be something to do with how records are collected from the database? Perhaps there is no ordering?

phalt:
@jinpark which page are you seeing this on? If we know where to start looking it'll help us discover a fix quicker! :grin:

jinpark:
For the seoul group, `https://djangogirls.org/seoul/applications/` page

I thought there was only one `/applications` page!

phalt:
@jinpark I've been investigate this, this evening and I can't seem to reproduce any weird state. Are you sure it is not just because you've set a filter at the top of the page?

phalt:
@jinpark I have a fix into master, should be soon so feel free to close this when you're happy!

helenst:
This is deployed now :) Looks good on Coventry!

jinpark:
LGTM here as well! Closing the issue! Thanks!!! :+1:

PR #119, opened 2015-08-10
Potentially fix random ordering in /applications/

phalt:
Solves #117. Orders by ID by default if no other ordering is present.


helenst:
Yay! Thanks :) :octopus:

PR #123, opened 2015-08-16
Patreon manager

bmispelon:
Added an admin app to manage our Patreon supporters.

It's still a bit rough around the edges but it should help me figure out who I need to send stuff to.

### DIFF
--- a/patreonmanager/admin.py
+++ b/patreonmanager/admin.py
@@ -1,0 +1,57 @@
+from django.contrib import admin
+from django.contrib import messages
+from django.core.urlresolvers import reverse
+from django.template.defaultfilters import linebreaksbr
+from django.utils.html import format_html
+from django.utils.translation import ugettext_lazy as _, ungettext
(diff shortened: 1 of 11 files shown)


patreonmanager/admin.py
```diff
@@ -0,0 +1,57 @@
...
+from django.core.urlresolvers import reverse
+from django.template.defaultfilters import linebreaksbr
+from django.utils.html import format_html
+from django.utils.translation import ugettext_lazy as _, ungettext
+
+from .models import Patron, Reward, Payment
+
+
+@admin.register(Patron)
+class PatronModelAdmin(admin.ModelAdmin):
```
olasitarska: I think in the whole Django Girls website we use just PatronAdmin, without Model. Should we make it consistent?

This is a nitpick, sorry! ;)

bmispelon: Fixed.

Issue #112, opened 2015-08-03
Add FAQ page to main site. 


williln:
This is newbie-friendly! 

1. Take the [Organier's FAQ](http://organize.djangogirls.org/faq/index.html), pull out the questions that aren't specific to organizing, and duplicate them on an /faq/ page for the main Django Girls site. 
2. Site should be linked from the bottom footer under "Django Girls." 
3. URL should be /pages/faq

amureki:
@williln maybe we should create an app for FAQ? Or just leave it as flatpage?

olasitarska:
@amureki I think it's better to have it as html in repository, because people can PR against it! :)

andrayantelo:
I think this is closed by #147

olasitarska:
Thanks!

PR #126, opened 2015-08-19
added FAQ page

amureki:
According to #112 
Added FAQ page with questions in html.
Just need to create new flatpage with "faq.html" template and "/faq/" url and it wiil be ready.


phalt:
Wanna add the view code for rendering this too? Then it'd be amazing! :grin:

amureki:
@phalt as @williln mentioned that it should be with "/pages/faq" url, I've added page as flatpage template, so admins should just create new FAQ flatpage in admin panel and select faq.html template.
Because there is default flatpages url dispatcher by slug, I don't think we should manually add url dispatcher and view for this - it will be excess. 
Or am I wrong?
It's not problem to me to make it as custom page with own view.

phalt:
@amureki ah righty. I had assumed we're writing a view for it. Flat page is the way forward. Terribly sorry.

amureki:
@phalt all ok, no problem. :) I'll check what else I can do here.
Btw, i've sent you an email about "HN archived post", can you check it when you have some free time? 
Thank you!

williln:
@amureki Is this still active? 

I ask because since we started this process, we've refined how we're doing these sorts of flat pages. We're not putting them in the admin anymore, so that the content for them is easier to be seen and the editing process is more transparent. The new /foundation page is an excellent example. So basically, @phalt is right and we are handling these pages with views and templates, as opposed to controlling them in the admin. 

Are you still interested in working on this @amureki? I'm sorry for leading you on a wild goose chase! We had some flat pages in the admin for awhile, but decided it was better to put the content for all pages in version control. Thanks! 🐍

amureki:
@williln sure, I'll check this as soon as I can and remake this page in your codebase style.
Btw, don't you guys want to use CBV's?

williln:
@amureki I'm not sure on the CBVs! I think our views are simple enough that it's worth it to keep them the way they are. Also, the tutorial doesn't teach CBVs, and we want to keep the project easy for attendees to contribute to, so we probably want to leave them as they are.

amureki:
https://github.com/DjangoGirls/djangogirls/pull/147

Issue #135, opened 2015-09-28
Add DB rollback functionality 


lightstrike:
When transition was made to PythonAnywhere from Heroku, we lost DB rollback functionality. Add again using [this guide](https://help.pythonanywhere.com/pages/RegularPostgresBackups).

olasitarska:
@lightstrike any progress on this? If you don't have time to work on this right now, I'd appreciate some guidance of what needs to be done :)

lightstrike:
Started working on it with @pieropalevsky last night: https://github.com/DjangoNYC/djangogirls/commit/acf7fcc208453c877ff880874aad755a8a656daa - this feels like the right approach but getting the command to restore the *right* backup is the tricky part. If you have any thoughts there, that would be helpful.

PR #137, opened 2015-09-29
command to create PSQL dump/upload to S3

lightstrike:
First stab to address issue https://github.com/DjangoGirls/djangogirls/issues/135 so there are consistent backups to S3. Haven't added any tests (not sure how for AWS...), open to suggestions there. 

Also haven't added a bash script to add as a cron job. If we went the absolute path approach (`/path/to/virtualenv/bin/python/ /path/to/djangogirls/manage.py backup_postgres_to_s3`) we might not need/want a .sh file. It's possible we'll hit a permissions issue on PythonAnywhere.

### DIFF
--- a/core/management/commands/backup_postgres_to_s3.py
+++ b/core/management/commands/backup_postgres_to_s3.py
@@ -1,0 +1,47 @@
+# -*- encoding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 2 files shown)


olasitarska:
This looks good to me, but I think it'd be better if you could merge & deploy it, so you can test it on PythonAnywhere too. Make sense? :)

olasitarska:
And thank you, you completely rock!

lightstrike:
:back: :up: :clock:!

Issue #138, opened 2015-10-08
Disable editing organizers field for organizers


patjouk:
If an organizer click on a name in the team list at the bottom of the admin page, they are kicked from the organizing team and replace by the people they clicked on.

To prevent any problems, they shouldn't have the right to edit this field.

olasitarska:
So what we need to do here, is a little bit of magic that I talked about in my DjangoCon talk about admin :) :sparkles: 

What we want to achieve is have the `organizers` field on `Event` model to be readonly if user who is editing doesn't have a `superuser` status. Fortunately, looking at the [code here](https://github.com/DjangoGirls/djangogirls/blob/master/core/admin.py#L33), we're already doing it for the `email` field. So all we need to do is modify line 33 to return both `email` and `organizers` in a tuple, not just `email`.

Once we're on it: the organizers field is getting quite tough to edit because of how long this list is. We could add a `filter_horizontal = ['organizers']` to `EventAdmin` to make it a bit easier. [Relevant docs, if you're curious](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.filter_horizontal)

Let me know if you need any help @patjouk! :sparkles:

PR #140, opened 2015-10-12
Organizers list is now read only for organizers.

patjouk:
Linked to issue https://github.com/DjangoGirls/djangogirls/issues/138.

I think it's `team` and not `organizers` that should be made read only. I also added `filter_horizontal` and it's working locally :smile_cat:


olasitarska:
Perfect! :)

olasitarska:
Thanks @patjouk! :spaghetti:

Issue #142, opened 2015-10-14
Event with no date set is considered past event


patjouk:
If an event has no defined date, it's considered as a past event.
One of the consequence is that this event is also a past event on the events webpage.

williln:
Problem will be in djangogirls/core/models.py, EventQuerySet model, past method. Need to retrieve events that have a date at all, and then that have a date LT datetime.now().

PR #148, opened 2015-10-27
Exclude events with no date from past events

andrayantelo:
Fixes https://github.com/DjangoGirls/djangogirls/issues/142

I am excluding events with no date from the past events. How do you test this? When I made an event it never showed up on the page no matter what I did because of authentication failure.


aniav:
Thank you! :tada:

Issue #122, opened 2015-08-14
Add "Applicant declined" as "status" option


williln:
I have an attendee on the waitlist who won't be able to attend even if picked, and since her status is "waitlist," I can't add her RSVP as "rejected." My only option is to "reject" her application, which isn't really accurate. Propose adding new status in addition to "Application accepted, rejected, etc." of "Applicant declined," for those situations where applicants decline the wait list, or need to rescind their applications before decisions are made. 

Probably newbie-friendly!

PR #149, opened 2015-10-27
Added applicant declined state.

andrayantelo:
The state applicant declined is for applicants who choose not to be on
the waitlist and are not going to attend the event.

Closes https://github.com/DjangoGirls/djangogirls/issues/122


patjouk:
Perfect :heart_eyes_cat:

PR #151, opened 2015-10-30
Serve a not found response if event page does not exist

### LINKED ISSUE #144: Error when visiting an application form of a page that doesn't exist
screen shot 2015-10-16 at 11 29 44 am

laidibug:
Fixes #144

When a url like this was accessed:
"/blah/applications" the app got an error because the event was not
found.


olasitarska:
YAY! Thank you so much!

PR #153, opened 2015-10-30
Fix the apply view when two forms are created

### LINKED ISSUE #143: Error when there is more than one form for a page
screen shot 2015-10-15 at 9 33 47 am

laidibug:
When two forms are created for an event page and you go to the
apply view you will get an error. Instead of using .get on the query
we can just grab the first one. It seems to me that creating two forms
is a mistake by the user and we just need to handle it gracefully.

Fixes #143

### DIFF
--- a/applications/views.py
+++ b/applications/views.py
@@ -24,7 +24,7 @@ def apply(request, city):
         )
 
     try:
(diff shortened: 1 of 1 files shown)


olasitarska:
Thank you! :) :tulip:

PR #155, opened 2015-11-05
Fixed permalink

OPersian:
Little but not the least: "Learn More" is the first thing a visitor may click on


phalt:
@olasitarska is this template still used? I can't find a reference to it.

patjouk:
Thanks! :rose:

Issue #124, opened 2015-08-17
Applications - CSV Download should match current queryset


helenst:
If you're viewing a subset of applicants and click download, it stlil downloads everybody. The downloaded list should match the current page's list.

sophiadavis:
I'm going to give this one a shot!

aniav:
That's done! :)

PR #169, opened 2015-11-07
Application CSV download should only download visible applications

sophiadavis:
Fix for https://github.com/DjangoGirls/djangogirls/issues/124


aniav:
Thank you! :tada:
I believe `unittest.mock` was introduced in python 3.3 so that's why it was giving errors for you probably. I will revert to it after merge. Awesome work! :)

Issue #146, opened 2015-10-21
CSV columns misalign if new questions are added to form


ryangallen:
<img width="618" alt="screen shot 2015-10-21 at 5 17 59 pm" src="https://cloud.githubusercontent.com/assets/1379681/10650488/d33b6b28-7817-11e5-83b8-56970d169a2a.png">

After adding the t-shirt size question, the applications without a t-shirt size get the Code of Conduct response in the t-shirt column.

olasitarska:
Thanks for reporting! Definitely valid bug :)

PR #170, opened 2015-11-07
add test for applications_csv download where new question added with …

toliwaga:
added test for issue #146 CSV columns misalign if new questions are added to form


aniav:
Thank you! :tada:

Issue #118, opened 2015-08-10
Application number should correlate to the url


jinpark:
For the seoul chapter at least
`https://djangogirls.org/seoul/applications/424` links to application number `110`

It would make more sense to have the url be `https://djangogirls.org/seoul/applications/110` since it would be easier to refer and link to

N.B. This might break a lot of existing urls out in the wild that are between teams so might need to do a site wide notification about this change when it gets deployed

RachellCalhoun:
Our team had some confusion over this. I hope the url and app number can be the same to eliminate unnecessary miscommunication.

hassanabidpk:
+ Also when you click on "Save and go to next" it takes you some random application. After 110, I would ideally want to be at 111.

helenst:
@jinpark this seems like a nice idea, url breakage will need to be handled but I think this would be good to have.

@hassanabidpk This should probably be a separate issue, but it seems like the "save and go to next" is intentionally random. Maybe it should be re-labelled to say it's random, and perhaps a separate button which takes you to the next in sequence, what do you think?

jinpark:
How about

Save & Go Previous | Save | Save & Go Next | Save & Go Random
------------ | ------------- | ------------ | -------------

olasitarska:
I think that's two options too many ;)

hassanabidpk:
Agree with @helenst  . Having two options is reasonable.

williln:
Also, it's in our blog post about how we score that apps are scored randomly, so no one is scored worse just because they are at the bottom of the pile. I don't think we should change that.

jinpark:
I'm ok with either way, I just want the application numbers/url sorted out :D

aniav:
So all in all this issue is about changing how the url is created. We should not change the behaviour of save & go to next. Summarising it up so we have a clear view (and because we have sprints on Django Under The Hood and I want this to be clear for anyone).

jinpark:
thanks @aniav ! I can have a PR for this fairly soon I think

aniav:
Awesome! Keeping my fingers crossed! :)

jinpark:
PR up for this one. Let me know if you have any problems :)

aniav:
Closing as it is now done!

PR #173, opened 2015-11-09
use app.number instead of id in url

jinpark:
For https://github.com/DjangoGirls/djangogirls/issues/118

Just changed the url so it looks for `app.number` instead of `app.id` 

I looked around but could't find any other places that it might have been referenced but I'm fairly sure that I might have missed something.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -218,6 +218,10 @@ class Application(models.Model):
     def is_accepted(self):
         return self.state == 'accepted'
 
+    @property
+    def city(self):
+        return self.form.page.event.city
+
     def __str__(self):
         return str(self.pk)
 
--- a/applications/models.py
+++ b/applications/models.py
@@ -218,10 +218,6 @@ class Application(models.Model):
     def is_accepted(self):
         return self.state == 'accepted'
 
-    @property
-    def city(self):
-        return self.form.page.event.city
-
     def __str__(self):
         return str(self.pk)
 
--- a/applications/views.py
+++ b/applications/views.py
@@ -82,7 +82,6 @@ def applications(request, city):
         'all_applications_count': Application.objects.filter(form__page=page).count(),
         'order': order,
         'menu': get_organiser_menu(city),
-        'city': city
     })
 
 @organiser_only
@@ -110,7 +109,7 @@ def application_detail(request, city, app_number):
     """
     Display the details of a single application.
     """
-    application = [app for app in Application.objects.filter(number=app_number) if app.city == city][0]
+    application = Application.objects.get(form__page__url=city, number=app_number)
     try:
         score = Score.objects.get(
             user=request.user, application=application)
--- a/applications/views.py
+++ b/applications/views.py
@@ -82,6 +82,7 @@ def applications(request, city):
         'all_applications_count': Application.objects.filter(form__page=page).count(),
         'order': order,
         'menu': get_organiser_menu(city),
+        'city': city
     })
 
 @organiser_only
@@ -105,11 +106,11 @@ def applications_csv(request, city):
 
 
 @organiser_only
-def application_detail(request, city, app_id):
+def application_detail(request, city, app_number):
     """
     Display the details of a single application.
     """
-    application = Application.objects.get(pk=app_id)
+    application = [app for app in Application.objects.filter(number=app_number) if app.city == city][0]
     try:
         score = Score.objects.get(
             user=request.user, application=application)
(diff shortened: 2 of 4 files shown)


applications/views.py
```diff
@@ -105,11 +106,11 @@ def applications_csv(request, city):
 
 
 @organiser_only
-def application_detail(request, city, app_id):
+def application_detail(request, city, app_number):
     """
     Display the details of a single application.
     """
-    application = Application.objects.get(pk=app_id)
+    application = [app for app in Application.objects.filter(number=app_number) if app.city == city][0]
```
helenst: Here, you can use the database will pick the right entry for you.

Suppose you request /yourcity/applications/1 - the filter query here will return application number 1 for every event that has one, and then your list comprehension will perform another database query for each one of those to check which city it is. So as Django Girls gets bigger and bigger, there will be more and more queries here!

It is usually best to let the database do this work! You can add more parameters to the filter query so you can query by both city and number.

Also, the field you have used in the `city` property above will not get the right result - this is the display name of the city, which is not the same as what you have in the URL. The correct field would be `app.form.page.url` (so you could do `.filter(form__page__url=city)`

templates/applications.html
```diff
@@ -57,7 +57,7 @@ <h3 class="page-header">Applications for {{ page.title }}</h3>
                   {% for app in applications %}
                   <tr>
                       <td><input type="checkbox" name="application" value="{{ app.id }}" /></td>
-                      <td><a href='{{ app.id }}'>{{ app.number }}</a></td>
+                      <td><a href='{{ app.number }}'>{{ app.number }}</a></td>
```
helenst: I know it was this way already, but we should really be using `{% url %}` for this.

templates/applications.html
```diff
@@ -109,7 +109,7 @@ <h3 class="page-header">Applications for {{ page.title }}</h3>
                           {% endif %}
                       </td>
                       <td class="text-right">
-                          <a href="{{ app.id }}" class="btn">View Application</a>
+                          <a href="{{ app.number }}" class="btn">View Application</a>
```
helenst: again, should use `{% url %}`

applications/views.py
```diff
@@ -82,6 +82,7 @@ def applications(request, city):
         'all_applications_count': Application.objects.filter(form__page=page).count(),
         'order': order,
         'menu': get_organiser_menu(city),
+        'city': city
```
helenst: I don't know if this needs to be passed in to the template.

applications/models.py
```diff
@@ -218,6 +218,10 @@ def get_by_rsvp_code(self, code, page):
     def is_accepted(self):
         return self.state == 'accepted'
 
+    @property
+    def city(self):
+        return self.form.page.event.city
+
```
helenst: This can be done with `filter` (see comment below)

applications/models.py
```diff
@@ -218,6 +218,10 @@ def get_by_rsvp_code(self, code, page):
     def is_accepted(self):
         return self.state == 'accepted'
 
+    @property
+    def city(self):
```
aniav: Adding properties that are using values from related tables should be done with special care. Calling this property will create additional database calls and we probably don't want that.

applications/views.py
aniav: I would suggest doing something like:
`Application.objects.get(number=app_number, form__page__event__city=city)`. Here you are iterating through all the applications retrieved from database and it will get extremely slow when we will have increasing number of applications and events, also we can just retrieve the very specific application from database with one query without the need for itarating in python. Additionally iterating through applications and calling app.city on every app object will produce a database query for each of them and we don't want that :)

I know this is some tricky stuff, you can catch me on slack if you would want to discuss it further ;)

aniav: Ah, I didn't see Helens comment! Sorry! :sparkles:

helenst:
Hi @jinpark, thank you so much for your work on this! I've made a few comments, please feel free to ask if they're unclear. Have a lovely day! :sunflower::whale:

applications/views.py
helenst: That is OK! It's good that we agree :smile:  :sparkling_heart:

jinpark: Ah, I didnt realize you can do `form__page__event__city=city` that nested!

I'll make the fix

templates/applications.html
jinpark: Agreed. I'll make the change!

jinpark:
I made the changes :) 

Let me know if I missed anything

aniav:
Thank you! :tulip:

PR #174, opened 2015-11-11
Added province and state to meetups and jobs form

### LINKED ISSUE #165: Add the State/Province field for meetups/jobs
Extracted for #162 

In /community/meetups, we're only displaying the city. As this page grows, we may want to display city and country, and possibly city/state-or-province/country. There are two Portlands in the US, for example. 

Probably the same for the Job model.

patjouk:
Fix #165 :heart:

### DIFF
--- a/jobs/admin.py
+++ b/jobs/admin.py
@@ -204,7 +204,7 @@ class PublishFlowModelAdmin(admin.ModelAdmin):
 class JobAdmin(PublishFlowModelAdmin):
     fieldsets = (
         ('Job info', {'fields': ('title', 'company', 'website', 'contact_email',
-                           ('cities', 'country'), 'description', 'remote_work',
+                           ('cities', 'country'), 'state_province', 'description', 'remote_work',
                            'relocation')}),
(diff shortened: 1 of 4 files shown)


jobs/migrations/0010_meetup_state_province.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
aniav: Could you squash the migrations in one? They seem to affect the same models ;)

applications/migrations/0015_auto_20151110_1054.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
aniav: Should this migration be in this Pull Request?

patjouk:
I removed the application migration (not sure why it was there!) and squashed the job-meetups' migrations in one :hammer: :tada:

olasitarska:
Thanks @patjouk!

Issue #156, opened 2015-11-06
Applications - Form - support apply as coach form


marteloge:
To make it easier for organizers it would be a nice feature to support applications from coaches in a form as we do with the attendents.

pguatura:
Hi! I've been looking for this feature and I'm very glad to see that someone already though about this and started developing it. I was thinking about continuing the development considering the feedback on the pullrequest, but I haven't contributed to anything else on this repository. Do you mind if I try or do you think it's better if i start with something else?

helenst:
Hi @pguatura I'm sorry nobody replied to your message on the pull request before. It's totally fine for you to take this on if you are happy working on it and it would still be a really great feature to have! There are probably smaller tickets available if you did want something easier to start with but you don't need to do that. Happy coding! :snake::sparkling_heart:

gconnect:
I would like to work on this issue #176 what do I need to do to get started?

PR #176, opened 2015-11-15
Coach form

peconia:
I added coach applications and messaging, similar to the applications for attendees but without the scoring. This could possibly fix issue #156. I added some unit tests and amended existing ones slightly so all the tests are passing. Everything works for me locally but I am not sure if this is good enough, please let me know what should be fixed and I'll do my best. Thanks! :scream: :tangerine: :hamster:


olasitarska:
Hi @peconia!

Thank you for working on this -- I see that this took a lot of effort!

I think it'd be much better if we just modify and refactor existing applications app to allow for multiple forms - this way organizers can create forms about whatever they want, and are not limited to just two forms. 

I might be wrong though, so happy to hear a second opinion from any @DjangoGirls/committers or @DjangoGirls/support?

helenst:
@olasitarska @peconia I agree that this is awesome work and it would be really great to be able to manage coaches more nicely! There is a lot of duplicated code between this and the applications module so it would be good to wrap it all up into a single codebase - having the ability to extend to more forms might be nice too :) :whale::sunflower:

peconia:
Oops, I kind of forgot this. If I ever find the time I'll have another go, but if anyone else wishes to fix it in the meantime then please don't wait for me. Thanks for the feedback :rabbit: :heart:

pguatura:
Hi! I've been looking for this feature and I'm very glad to see that someone already though about this and started developing it. I was thinking about continuing the development considering the feedback, but I haven't contributed to anything else on this repository. Do you mind if I try or do you think it's better if i start with something else?

PR #177, opened 2015-11-17
Fixed #160

### LINKED ISSUE #160: Add job/meetup forms could be formatted better
_From @MarysiaLowas on July 24, 2015 19:1_

For now, all input fields in Add job/meetup forms are set to 260px but it would be nicer to have them wider. And we could also improve the forms layout but we would need to dive into crispy forms.

_Copied from original issue: djangogirlsjobs/djangogirls#68_

MarysiaLowas:
Removed crispy forms from the project, that affected the following forms:
- add job
- add meetup
- password reset
Tweaked community CSS for better formatting (at least I hope they look a bit better now)
Created a custom template filter "addclass" that adds a CSS class to a form field. That was done for more Boostrap-like form styling.

### DIFF
--- a/jobs/templatetags/jobs_tags.py
+++ b/jobs/templatetags/jobs_tags.py
@@ -1,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+@register.filter(name='addclass')
+def addclass(field, new_classes):
+
+    old_classes = field.field.widget.attrs.get('class', '')
+    classes = ' '.join([old_classes, new_classes])
+    return field.as_widget(attrs={"class": classes})
--- a/djangogirls/settings.py
+++ b/djangogirls/settings.py
@@ -39,7 +39,6 @@ INSTALLED_APPS = (
     'easy_thumbnails',
 
     'django_countries',
-    'crispy_forms',
     'bootstrap3_datetime',
 
     'ckeditor',
@@ -173,7 +172,6 @@ MAPBOX_MAP_ID = 'olasitarska.m8nged0f'
 
 APPEND_SLASH = True
 
-CRISPY_TEMPLATE_PACK = 'bootstrap3'
 CKEDITOR_UPLOAD_PATH = "uploads/"
 CKEDITOR_CONFIGS = {
     'default': {
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,6 @@ cssselect==0.9.1
 dj-database-url==0.3.0
 django-ckeditor==5.0.2
 django-countries==3.4.1
-django-crispy-forms==1.5.2
 django-date-extensions==1.1.0
 django-leaflet==0.16.0
 django-bootstrap3-datetimepicker==2.2.3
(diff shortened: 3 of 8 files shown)


olasitarska:
@aniav, @rebkwok or @ksiedlarek would you have some time soon to review this? I think you're more familiar with the jobs/meetups code.

Otherwise I can schedule some time next week :)

jobs/templatetags/jobs_tags.py
```diff
@@ -0,0 +1,7 @@
+from django import template
+
+register = template.Library()
+
+@register.filter(name='addclass')
+def addclass(value, arg):
+    return value.as_widget(attrs={'class': arg})
```
rebkwok: Would this be better added with the form creation?  Kind of torn - this is more DRY since you only have to use it once in the form for all fields, but then again, it's custom code for something django already does.  You could do something like this in the form to set the widget attrs for all fields http://stackoverflow.com/a/3800247
If you do keep with the template filter, I think addclass will override any existing 'class' attr that may have been set on a field's widget in the form creation.  This might not be what you want (if it is, maybe the filter would be better named something like 'set_css_classes'?)

MarysiaLowas:
I'm sorry for the late comment, there were a few other things going on...

Well, the reason I've done this with the template filter rather than in the form creation, was to keep the HTML and CSS code outside of Python, as this was one of @aniav objections to crispy forms. I based it quite heavily on this post:
http://vanderwijk.info/blog/adding-css-classes-formfields-in-django-templates/

There is a comment to this post to make the filter add a CSS class instead of overwriting it:

> @register.filter(name='addcss')
> def addcss(field, css):
> class_old = field.field.widget.attrs.get('class', None)
> class_new = class_old + ' ' + css if class_old else css
> return field.as_widget(attrs={"class": class_new})

Another solution would be to use this library: https://pypi.python.org/pypi/django-widget-tweaks
But, I'm not sure whether it is a good idea to add another library just to use it for adding one class to three forms.

Or I can add this class in the form creation as @rebkwok suggested.

I really don't know which solution is the best, so please let me know what you think.

aniav:
I stringly believe in decoupling of frontend and backend stuff. I would stay with `addcss` but I would also reuse the classes already present on the widget to something like the code from comment posted by @MarysiaLowas above but I would also clean it up a little bit to something like this:

```python
@register.filter(name='addcss')
def addcss(field, new_classes):
    old_classes = field.field.widget.attrs.get('class', '')
    classes = ' '.join(old_classes, new_classes)
    return field.as_widget(attrs={"class": classes})
```

MarysiaLowas:
@aniav @rebkwok @ksiedlarek @olasitarska I pushed the update some days ago. Anyone to have a look?

aniav:
I promise to take a look at it in the next few days!

olasitarska:
@aniav will you have time for this next week? otherwise, please assign to me, I can take a look! :)

aniav:
@olasitarska perfect timing, I am looking at it right now. Finally found time! :D

olasitarska:
Awesome :)

aniav:
Everything looks good! Thank you @MarysiaLowas and sorry it took so long! :sparkles:

PR #187, opened 2015-12-03
Add soft delete for event and website

mmoravcik:


core/models.py
```diff
@@ -87,8 +90,10 @@ class Event(models.Model):
     main_organizer = models.ForeignKey(User, null=True, blank=True, related_name="main_organizer")
     team = models.ManyToManyField(User, blank=True)
     is_on_homepage = models.BooleanField(default=False)
+    is_deleted = models.BooleanField(default=False)
 
-    objects = EventQuerySet.as_manager()
+    objects = EventManager()
+    all_objects = models.Manager()  # This includes deleted objects
```
olasitarska: Could we call this `with_deleted`? Otherwise we've got Event.objects.all() and Event.all_objects.all(), which is a bit weird :)

core/models.py
```diff
@@ -143,6 +157,10 @@ class EventPage(models.Model):
     url = models.CharField(max_length=200, null=True, blank=True)
 
     is_live = models.BooleanField(null=False, blank=False, default=False)
+    is_deleted = models.BooleanField(default=False)
+
+    objects = EventPageManager()
+    all_objects = models.Manager()  # This includes deleted objects
```
olasitarska: Could we call this `with_deleted`? Otherwise we've got EventPage.objects.all() and EventPage.all_objects.all(), which is a bit weird :)

olasitarska:
Two small nitpicks here. You can update this pull request by pushing to this branch.

Otherwise it looks great, thank you!

olasitarska:
Ok, I haven't realized that `future` and `past` things are going to be more complicated with that. Merging then!

PR #188, opened 2015-12-03
185 contact form

mmoravcik:
It will create copy of the email in ContactEmail model.

I moved around some CSS and created common.styl, so please double-check the styling still works across various sections

Also, check and update my wording :smiley_cat:


olasitarska:
Great job! So helpful. :sparkles: 

I've got one suggestion -- maybe we could make ContactForm a ModelForm and move the email sending logic from the views.py to save() of this ContactForm? I think that would be a lot more neat to keep it there. Let me know what you think :)

mmoravcik:
You probably mean the new `ContactEmail` model I created for saving the sent email.

Initially I wanted just to save the message and sender, but at the end it look like the model is mirroring the `ContactForm`.

I will change it to Modelform

mmoravcik:
@olasitarska have a look now :christmas_tree:

mmoravcik:
@olasitarska have a look now :christmas_tree:

core/models.py
```diff
@@ -174,6 +175,58 @@ def delete(self):
...
+            )
+        except SMTPException:
+            self.sent_successfully = False
+
+        super().save(*args, **kwargs)
+
+    def _get_to_email(self):
+        if self.event and self.event.email:
+            return self.event.email
+        return 'hello@djangogirls.com'
```
olasitarska: Should be hello@djangogirls.org

olasitarska:
This looks excellent now! I'll test that locally in an hour :)

olasitarska:
Thanks @mmoravcik, this is awesome! :sparkles:

PR #189, opened 2015-12-03
Add Django 1.9 support WIP

mmoravcik:
DO NOT MERGE YET

This is WIP

Waiting for model mommy authors to submit fix for 1.9:
https://github.com/vandersonmota/model_mommy/issues/262

Also, django-bootstrap3-datetimepicker, pointing to specific commit for now, so we need to keep an eye on the actual release:
https://github.com/nkunihiko/django-bootstrap3-datetimepicker/pull/35

UPDATE:
model mommy updated, 1 test is failing though https://travis-ci.org/DjangoGirls/djangogirls/builds/95775351


olasitarska:
@MarysiaLowas this test is failing on the last line: 
```
def test_unassigning_job_reviewer_from_under_review_post(self):
        unassign_reviewer_url = reverse(
            'admin:unassign_job_reviewer',
            args=[self.job_under_review.id]
        )
        response = self.client.get(unassign_reviewer_url, follow=True)
        self.assertEqual(response.status_code, 200)
        self.assertIn("Open", str(response.content))
        self.assertIn("None", str(response.content))
```

Do you remember what was that last line testing? I think Django changed the default option in the select from "None" to --- in Django 1.9, but I'm not exactly sure if this is causing it. 

We could also just remove this line from the test, but I'm not exactly sure what's the purpose of it :) 

Thanks!

MarysiaLowas:
@olasitarska, the 'unassign' action should clear the reviewer field by setting the reviewer to 'None' and this is what is shown in the admin. So the last line was supposed to test that. Probably not the best way to do that... ;)

olasitarska:
Ok, I'm reviewing this now.. Will fix this failing test! :)

olasitarska:
@MarysiaLowas if you're wondering how I fixed that test: https://github.com/DjangoGirls/djangogirls/commit/4323fcc49593212feb8871f1dfa9c7050ab7ba80 :) 

Thanks @mmoravcik! :hand:

MarysiaLowas:
Thanks @olasitarska :heartbeat:

PR #190, opened 2015-12-04
first attemp for FAQ page

patjouk:
:warning: not to merge right now: waiting for contact form first!:warning: 
Those are the recurring questions I received at hello@. Did I forget something?


olasitarska:
@patjouk Looks good Lucie!

I would improve two things:
- replace hardcoded url to contact page with reversing URL
- maybe we could move this bit of a faq into separate template, and include it both here and above the contact form? I think people won't click on the link to FAQ on contact form.. This way we still have to update content once, but we're displaying it in both places :)

patjouk:
I'll change the links :wink: 
I totally share your opinion about moving stuff above the contact form. I'll do that!

Issue #179, opened 2015-11-22
Sharing event sites to facebook


helenst:
It would be nice if, when sharing an event website to Facebook, it could show information for the event itself instead of the default Django Girls metadata and image.
e.g.
![screenshot-www facebook com 2015-11-22 14-56-01](https://cloud.githubusercontent.com/assets/217966/11324414/2c3b09ac-9129-11e5-867d-7bded11dcf7a.png)

I think this involves customising the opengraph tags in the header.

ulgens:
@helenst Is that issue resolved?

helenst:
I think it needs more opengraph tags - I think it is still using the main site's description - it'd be nice to use `og:description` and have the description of the relevant event, and also good to use the city image rather than default logo perhaps?

helenst:
Closing as I did this yesterday (https://github.com/DjangoGirls/djangogirls/commit/8170a040a9b35011db341f24a42b7ffe14c72e10).

PR #191, opened 2015-12-05
changed og:url to point to the city's event url rather than main domain on the base event template

peconia:
When sharing the link to the event page on Facebook the link points to the main django girls domain rather than to the specific event. I changed it to include the city so that you can link directly to the event.


helenst:
Yay, thank you! Fixes #179

peconia:
Thanks! I didn't realise it had an issue open :) :coffee: :fast_forward: :santa:

PR #193, opened 2015-12-07
FAQ page updated and main questions added to contact page.

patjouk:
:tada:

### DIFF
--- a/templates/contact.html
+++ b/templates/contact.html
@@ -9,11 +9,14 @@
 
                 {% include 'includes/_info_messages.html' %}
 
+                <p>Hi! We would love to hear about you!</p>
+
+                <p>To be sure your email reach the good person, check this FAQ first!</p>
+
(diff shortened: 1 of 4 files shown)


templates/contact.html
```diff
@@ -9,11 +9,14 @@
 
                 {% include 'includes/_info_messages.html' %}
 
+                <p>Hi! We would love to hear about you!</p>
```
olasitarska: I think it should be "hear from you"

templates/contact.html
```diff
@@ -9,11 +9,14 @@
...
+
+                <p>To be sure your email reach the good person, check this FAQ first!</p>
```
olasitarska: *the right person

olasitarska:
@patjouk fixed those two typos and added some styles. thanks! :) :dolphin:

PR #194, opened 2015-12-08
Linked event with no website to temporary webpage

patjouk:
I realized that events with no website have a nice [temporary page](https://github.com/DjangoGirls/djangogirls/blob/master/templates/event_not_live.html) but it was not used. It will probably help to reduce emails received at hello@ :wink:


olasitarska:
Awesome, thanks @patjouk! Could you deploy this?

patjouk:
done :wink:

PR #195, opened 2015-12-09
Add a new command to manage.py to sync trello events board.

patjouk:
To keep an eye on every event planned, I created a Trello board a few months ago. Today, I finally had the time to update my old sync script and to add it as a new command for manage.py. :sparkles: 
Of course, I'll write some docs for the wiki :wink:


olasitarska:
Yay, thanks @patjouk!

Issue #159, opened 2015-11-07
Add "Add another job" button to submission page


aniav:
_From @williln on July 15, 2015 19:34_

On the success page, I think it'd be nice to have an "Add another job" button. A company may have several opportunities. 

_Copied from original issue: djangogirlsjobs/djangogirls#52_

ulgens:
@aniav Is the task completed?

aniav:
@ulgens seems so! Thank you, I am closing it :)

PR #196, opened 2015-12-15
Add 'Add another job' button to submission page #159

cgl:
Add 'Add another job' button to submission page #159


olasitarska:
Thanks @cgl! :tulip:

PR #201, opened 2016-01-15
Made it easier to find event when adding postmortems

patjouk:
We now have so many events (:heart_eyes_cat: by the way) that it became difficult to find the event you're looking for in the dropdown menu. Changing it for a search field will probably make our life easier :wink:


olasitarska:
YAY!

PR #202, opened 2016-01-15
Event date field can't be empty or a year only

patjouk:
When we create an new event, it has to be at least a month and a year. Yet, organizers can change it to only a year, which make it really difficult for us to know when it will really happen (we don't want and don't have the time to send them an email every month to see if they have a more precise date or not :wink:).
I also made it impossible for organizers to not give a date to their event for the same reason.


olasitarska:
Woohoo! :boom:

PR #203, opened 2016-01-15
Add two items in setup email

patjouk:
Added three things in setup email send to organizers:

* Offering the opportunity to announce that registration are opened in the Dispatch.
* How to do if they have to postpone or cancel an event. Few events didn't happen and maybe adding that we won't be angry at them in this email will convince organizers to contact us and not just disappear.
 * Link to FAQ :wink:.


olasitarska:
Thanks @patjouk!

PR #204, opened 2016-01-20
Fixed a bug where two applications could have the same number in the database

patjouk:
Rewrote the save function to be sure that 2 applications can't get the same number in the database ([link to opbeat](https://opbeat.com/django-girls/djangogirlsorg/errors/18/))


olasitarska:
That should work! Thanks :)

Issue #206, opened 2016-01-20
Don't include declined applications when scoring.


andyosuna:
I noticed that after declining an application you can still score it, maybe adding a "skip" button or not including them in the query at all would fix this.

aniav:
It seems this has been fixed already :)

PR #208, opened 2016-01-20
duplicate applications and declined scoring

### LINKED ISSUE #205: Prevent duplicate application entries
Add a unique on the email field in the applications to prevent people from applying more than once.

andyosuna:
Prevent people from submitting multiple applications - #205 
Prevent scoring declined applications - #206

### DIFF
--- a/applications/forms.py
+++ b/applications/forms.py
@@ -17,7 +17,6 @@ class ApplicationForm(forms.Form):
         super(ApplicationForm, self).__init__(*args, **kwargs)
         self.fields = generate_form_from_questions(questions)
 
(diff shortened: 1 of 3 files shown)


olasitarska:
I don't understand how this fixes the duplicate emails problem to be honest -- I think this is just going to throw an error for anyone who has application to any event already?

A test might be useful to solve that.

The other fix for excluding rejected applications is correct, but I can't merge it together with that other yet.

olasitarska:
@andyosuna please re-open whenever you've got time to work on this again! :)

PR #213, opened 2016-01-22
Replacing Simple Captcha by reCAPTCHA

patjouk:
Captcha made by simple Captcha are not easy to read at all: let's change it for reCAPTCHA :smile: 
Everything is already configured on PythonAnywhere :wink:

### DIFF
--- a/djangogirls/settings.py
+++ b/djangogirls/settings.py
@@ -154,6 +154,11 @@ DEFAULT_FROM_EMAIL = "hello@djangogirls.org"
 
 SLACK_API_KEY = os.environ.get('SLACK_API_KEY')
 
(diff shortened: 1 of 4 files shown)


djangogirls/settings.py
```diff
@@ -154,6 +154,11 @@
 
 SLACK_API_KEY = os.environ.get('SLACK_API_KEY')
 
+RECAPTCHA_PUBLIC_KEY = os.environ.get('RECAPTCHA_PUBLIC_KEY ')
```
olasitarska: Is that gonna work with this one space at the end?

patjouk: oups ><'

olasitarska:
Awesome! Thanks @patjouk

Issue #214, opened 2016-01-25
Delete question in form not working


patjouk:
When an organizer try to delete a question in their form, they get a 404.
Seems to be a URL problem (https://djangogirls.org/admin/applications/question/*number*/change/delete/ is not working but https://djangogirls.org/admin/applications/question/1035/delete/ is).

patjouk:
fixed by #215

PR #215, opened 2016-01-25
Bump django-suit version.

alexanderad:
This addresses https://github.com/DjangoGirls/djangogirls/issues/214

It appears to be compatibility issue with Django 1.9 which has been fixed in https://github.com/darklow/django-suit/commit/2973aa5e53b7b4c947bf3d51b914feaa974d4512


aniav:
Thank you! :)

PR #216, opened 2016-01-25
Tiny fix for 2.x compatibility issues.

alexanderad:
While setting up locally my env I faced two small issues (because I use python 2.7 in here):
* models decorated with `@python_2_unicode_compatible`  should have a single `__str__` method declared (otherwise an error raised during model load) [1]
* 2.7 does not understand short form of `super` call (I see all elsewhere in project only full form call is used)

[1] https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.encoding.python_2_unicode_compatible

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -205,7 +205,7 @@ class ContactEmail(models.Model):
     class Meta:
         ordering = ('-created_at',)
 
-    def __unicode__(self):
+    def __str__(self):
         return "%s to %s" % (self.email, self.sent_to)
 
     def save(self, *args, **kwargs):
@@ -222,7 +222,7 @@ class ContactEmail(models.Model):
         except SMTPException:
             self.sent_successfully = False
 
-        super().save(*args, **kwargs)
+        super(ContactEmail, self).save(*args, **kwargs)
 
     def _get_to_email(self):
         if self.event and self.event.email:


phalt:
@alexanderad I think that because we are running Py3 on djangogirls.org, these changes will break on the live site? @olasitarska or @bmispelon know more about the set up.

alexanderad:
@phalt I don't think they break, this just makes 2.7 comfortable as well.

aniav:
@alexanderad according to: http://stackoverflow.com/questions/1307014/python-str-versus-unicode we should have `__str__` as well as `__unicode__` not one or another :)

alexanderad:
Hmm, @aniav, I'm looking at official django docs over here: https://docs.djangoproject.com/en/dev/topics/python3/#str-and-unicode-methods and it says, that if model is decorated with `@python_2_unicode_compatible` (ContactEmail is decorated), defining a single `__str__` will handle all the cases both for Py3 and Py2. Beside that all other models do declare only `__str__`.

Anyways, I'm not insisting on this being merged, this is something I faced while setting up project locally:
```python
djangogirls [master] ./manage.py
/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/utils/six.py:808: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  return meta(name, bases, d)

Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 327, in execute
    django.setup()
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/apps/registry.py", line 108, in populate
    app_config.import_models(all_models)
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/apps/config.py", line 202, in import_models
    self.models_module = import_module(models_module_name)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/Users/darednaxella/Code/Python/djangogirls/core/models.py", line 184, in <module>
    class ContactEmail(models.Model):
  File "/Users/darednaxella/Code/Python/djangogirls/.venv/lib/python2.7/site-packages/django/utils/six.py", line 840, in python_2_unicode_compatible
    klass.__name__)
ValueError: @python_2_unicode_compatible cannot be applied to ContactEmail because it doesn't define __str__().
```

aniav:
Ah, ok, didn't notice the part with `python_2_unicode_compatible`. You are totally right. Thank you! :)

PR #217, opened 2016-01-28
responsive sponsor images

RachellCalhoun:
Python, python anywhere images were off screen on mobile. This should keep them in place.


olasitarska:
Thanks @RachellCalhoun! :dancer:

PR #219, opened 2016-01-30
Add preventDefault() so that when applications' state is changed, the…

sodevious:
On the applications page, when a user changes the state of the application, the page jumps to the top. This can be frustrating with events with many applications, for example NYC has 100+ applications. This commit should fix that behavior.

:) ✨


olasitarska:
Thanks @sodevious! :dolphin:

PR #221, opened 2016-02-01
Code editor

patjouk:
Editing code in a text field is not really fun and tend to produce errors (non closed html tags for example).
I changed this text field for a code editor (codemirror): it was nice but the code editor window was not resizable and was recovering the right menu of Django Suit.
So... long story short, I used JQuery UI to make this window resizable.
It wouldn't have been possible without @madjar's help, so thanks a lot :heart_eyes_cat:


olasitarska:
That is AWESOME! :) :dolphin: :dancer:

PR #223, opened 2016-02-03
Captcha

patjouk:
Two little things in this PR:
- Captcha added to contact form
- On newsletter page, previous dispatches were listed from the oldest to the newest: it's probably better to have the latest first ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ include
 .venv
 static/media
 .idea/*
+.cache/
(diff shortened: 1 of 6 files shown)


olasitarska:
On testing locally, I get this:

screenshot 2016-02-08 09 28 00

Is that expected?

patjouk:
Because of how Recaptcha works, the widget will always show an error in local mode but now, it will be possible to save a form in local, even if it's still complaining :wink:

olasitarska:
Excellent! :)

olasitarska:
Thanks! :panda_face:

PR #224, opened 2016-02-07
made adjustments to event template text

kerstin:
As discussed a while ago on Slack, I sat down to modify (part of) the template text for events.

In particular I changed instances of "female" to woman/women and removed references to the EuroPython conference as they are not applicable to most DG workshops. I also introduced some grammar & style fixes and added placeholder text for event-specific dates, numbers, places.

### DIFF
--- a/templates/default/values.html
+++ b/templates/default/values.html
@@ -2,47 +2,50 @@
 	<div class="col-md-6">
 		<h3>Django Girls</h3>
 		<p>
-			If you are a female and you want to learn how to make websites, 
-			we have good news for you! We are holding a one-day workshop for beginners!
+			If you are a woman and want to learn how to make websites, we have
+			good news for you: we are holding a one-day workshop for beginners!
 	    </p>
 
 		<p>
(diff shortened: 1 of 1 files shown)


templates/default/values.html
```diff
@@ -2,47 +2,50 @@
...
 
 	 <div class="col-md-6">
-	 	<h3>Apply for a pass!</h3>
+	 	<h3>Apply for the workshop!</h3>
 	 	<p>
-	 		If you are a woman, you know English and have a laptop - you can apply
-	 		for a pass! You don’t need to know any technical stuff - workshops
-	 		are for people who are new to programming.
+	 		If you are a woman, know English and have a laptop, you can apply
+	 		for our event! You dont need to know any technical stuff &ndash; our
```
olasitarska: should don't not dont

kerstin: Yup, that angled apostrophe seems to have got lost when I was trying to replace it. Fixed now.

olasitarska:
Thank you! :dolphin:

PR #227, opened 2016-02-09
Small corrections in setup email and applicants form

patjouk:
Two small correction:
- The dispatch is now send once every two weeks.
- Point 3 of setup email for organizers is not useful anymore since we can now filter events with open registration.
:tada:


olasitarska:
Yay! Instamerge.

olasitarska:
@patjouk could you deploy? :)

PR #229, opened 2016-02-12
Added links to FAQ, Organizers Google Group and Slack.

patjouk:
Let's make even easier for organizer to find the FAQ, organizer google group and slack channel!  
I'll add this to the organizers manual :wink:


olasitarska:
YAY!

Issue #231, opened 2016-02-15
Application Form: Questions - Add filter by form


cgl:
A filter by form option would make it easier to organize the Questions in the admin. 
I propose to add a filter on the form field so that we can distinguish the questions for a specific event from the others easily.

patjouk:
Fixed by #232 Thanks @cgl :cake:

PR #232, opened 2016-02-15
 Application Form: Questions - Add filter by form #231

cgl:
A filter by form option would make it easier to organize the Questions in the admin.
In this branch I added a filter on the form field so that we can distinguish the questions for a specific event from the others easily.


olasitarska:
Yay!

cgl:
Great, is it already deployed?
I still cannot see the filtering in my own admin page https://djangogirls.org/admin/applications/question/

helenst:
@cgl It wasn't, but I've just deployed it so you should see the filter now! :dolphin:

cgl:
@helenst thanks :tada:

PR #233, opened 2016-02-15
Country count fix

patjouk:
This should fix the problem :wink:


core/views.py
```diff
@@ -21,6 +21,7 @@ def index(request):
         'patreon_stats': FundraisingStatus.objects.all().first(),
         'organizers_count': User.objects.all().count(),
         'cities_count': Event.objects.values('city').distinct().count(),
+        'country_count': Event.objects.values('country').distinct().count,
```
olasitarska: shouldn't be count()?

patjouk:
:disappointed: argh! It's fix now!

olasitarska:
Thanks! :)

PR #235, opened 2016-02-21
Applications: add 'selected' counter, preserve ordering.

alexanderad:
While working with applications I though it would be nice to have two simple features on Applications page:
* ordering which is preserved when selecting status
* selected applications counter:


alexanderad:
The reason for that is me in need to select exactly 45 out of 175 :-)

olasitarska:
Thanks Alex, this is so cool!

PR #237, opened 2016-02-23
Fix layout

calculation:
Fixes two issues that break the layout of the home page on smaller screens. The vimeo iframe has a fixed width and breaks out of its container, and the button text does not wrap, resulting in the same issue. Tested on firefox 44 and safari 9.


aniav:
Thanks! :)

PR #238, opened 2016-02-26
remove fixed height from footer css

calculation:
The footer has a fixed height, causing the contents to overflow in some situations. This removes `height` from `footer` and `.page-wrap:after`, and `margin-bottom` from `.page-wrap`.

### DIFF
--- a/static/css/global.css
+++ b/static/css/global.css
@@ -505,15 +505,10 @@ span.highlighted-number {
 */
 .page-wrap {
   min-height: 100%;
-  margin-bottom: -300px;
 }
 .page-wrap:after {
   content: "";
   display: block;
-}
-footer,
-.page-wrap:after {
-  height: 300px;
 }
 footer {
   background: #f9f7f7;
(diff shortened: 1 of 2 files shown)


olasitarska:
Thanks for the contribution! Could you show an example of when the content overflows?

As far as I remember, the height was there to make sure that footer always sticks to the bottom, even if content of the page is too short. Is that still working correctly after your changes?

calculation:
Hi @olasitarska, the footer overflows when the width of the viewport is < 992px, and between 1025 and 1199px. The background of the footer doesn't extend to the bottom of the page, so it should be fairly easy to spot. I can post some images if you're unable to reproduce the issue.

Regarding your concern, I think the correct way to address this is by specifying a `min-height` on the output of `{% block content %}`. This ensures the footer won't be floating in the middle of the page, and eliminates the need to specify negative margins and fixed heights, as well as the related code maintainability issues.

calculation:
@olasitarska yes, I plan to send changes within a few days, and I'll test to ensure the footer is positioned properly.

olasitarska:
@calculation Please re-open the pull request once the changes are made. Thanks!

PR #241, opened 2016-03-04
prefetch_related brokes the csv download if order_by('average_score')

cgl:
In the applications page (https://djangogirls.org/istanbul/applications ) when you sort the applications by score ( https://djangogirls.org/istanbul/applications/?order=-average_score ) and then try to download as csv (https://djangogirls.org/istanbul/applications/download/?order=-average_score ) you cannot.

The problem is if the order field is average_score we cannot order it via order_by since it is only a property. Sorted is used and a list instead of a QuerySet is returned by function get_applications_for_page. That is all good.   
The problem starts when this returned list or QuerySet object is asked for prefetch_related('answer_set')

Do we really need the prefetch_related here? 

If the answer is no I propose to delete it. 

If the answer is yes let's just prefetch whenever possible as in https://github.com/cgl/djangogirls/tree/alternative_prefetch

We could use the sorted version of csv since it includes the stdev, so let's solve this little bug :)

Thanks,


helenst:
Thank you for your contribution! :)

I've been having a look at this. I see a few problems with the original code...

1. The blanket `except` is masking errors, and this is bad generally
2. The exception that's being hidden is because `get_applications_for_page` sometimes returns a QuerySet and sometimes a list. It returns a list when there's an ordered clause because we're using sorted() rather than proper queryset sorging.
3. The prefetch_related is needed because the answers are being accessed in the CSV request - it's not needed for the regular listing, which is why there isn't a problem there.

This fix would make the download work in all cases, but would make all the other CSV downloads are slow due to excessive querying. The alternative fix you proposed would do better than that but your ordered by downloads would still be slow (and I still don't like the blanket exception catching or the fallback approach to handling the return value)

I think the answer is either to make `get_applications_for_page` return a QuerySet in all cases, or to bring the `prefetch_related` inside that function so it can be applied to the queryset before the sorting happens.

It would be great to get a second opinion about this though! :tulip::turtle:

cgl:
@helenst thank you for your comments. I agree with your points.

About the original code: 
1) The same `except` used in line 83 too and there are no comments or logging related.
2) The order by `average_score` or `-average_score` options are working ok while viewing but not in CSV request. We're using sorted() rather than proper queryset sorting because we want to take into account not only the average_score but also the standard deviation in this sorting. (it is not a field but a model function)
3) The `prefetch_related` causes CSV request to fail if order by average_score or -average_score requested.

I think bringing the `prefetch_related` inside `get_applications_for_page` function is a good idea. But the function is also called on line 84 (applications.view) which does not require a prefetch.

Let's hear a second opinion as you said, I can do the changes once we decide, or I can leave it to you, too :) 

It would be really nice to fix ordered csv download.

Cheers,

helenst:
Ah great, yes that other case should be fixed too, with caution of course to make sure the right exceptions are still being caught. The average score stuff might be tricky with pure querying I think - there is a StdDev aggregate function so might be possible with the right queryset magic but I guess bringing `prefetch_related` inside is probably easier. Yes it's not needed in both cases. We could either just let it happen anyway - it's only one extra query - or pass in a prefetch_answers flag to control whether or not it happens.

I'll leave it to you to make the changes if you're happy doing it! :sparkles:

helenst:
Hey @cgl do you think you will get a chance to work on this? I'd be happy to help if you're unsure about anything! :sunglasses:

cgl:
Hey @helenst , I have updated my branch and moved `prefetch_related` inside the get_applications_for_page function. 

Cvs download after sort (?order=-average_score ) works ok now.

applications/views.py
```diff
@@ -106,12 +106,8 @@ def applications_csv(request, city):
     rsvp_status = request.GET.getlist('rsvp_status', None)
     page = get_event_page(city, request.user.is_authenticated(), False)
     order = request.GET.get('order', None)
-    try:
-        applications = get_applications_for_page(
-            page, state, rsvp_status, order).prefetch_related('answer_set')
-    except:
-        return redirect('core:event', city=city)
-
+    applications = get_applications_for_page(
```
helenst: We should still have exception handling here, - `Form.DoesNotExist` is raised by the utility function so that should still be handled here.

helenst:
@cgl Thank you for making these changes. Just one comment and then I think it's fine :)

helenst:
Hey @cgl do you think you will get the chance to work on this? It'd be great to get it merged! If not, I'm happy to take a look and help get it finished :palm_tree::panda_face:

cgl:
@helenst 👍 
Could you please check again?

helenst:
Thank you! :tulip:

PR #259, opened 2016-04-03
exploratory work for event organizers admin view

lightstrike:
@olasitarska Opening attempts, once the right plumbing is in place we can write the actual logic :smile_cat:


olasitarska:
Opened #297 now.

PR #260, opened 2016-04-03
Preventing organizers to create more than one form

patjouk:
When organizers create more than one form in the applications app, it makes "sending messages to attendees" crash: more than one object gets returned.


olasitarska:
Thanks Lucie! :)

PR #261, opened 2016-04-03
Fix #192: Update README:  how to set up a PostgreSQL database

### LINKED ISSUE #192: Update readme to include instructions on how to create PostgreSQL database
The settings have been changed to use PostgreSQL instead of SQLite. It would be helpful to add instructions on how to install and create a database locally as using PostgreSQL is not as straightforward as SQLite.

collateraldamadge:
Updated README:  "Setting up a development environment" section now includes instructions on how to set up a local PostgreSQL database.

Thanks a lot for the project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,22 @@ Create a new virtual environment if needed. Then, install all the required depen
 
     pip install -r requirements.txt
 
-Create your database:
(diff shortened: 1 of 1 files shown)


olasitarska:
Thanks!

Issue #256, opened 2016-04-03
After setting applicant status, counter is not reset


helenst:
When you click checkboxes to select applicants on the Applications page, the number selected is shown at the top of the page.

When you choose the status and click "Apply", the statuses are changed and the applicant check boxes are deselected but the counter is not returned to zero.

This is not a big problem as the counter resets when you start clicking boxes again, but it might make a nice quick fix for somebody! :zap:

PR #270, opened 2016-04-23
Fixed a minor bug #256 so that number of selected applications get re…

jitendraag:
…set to 0 once a status change is "Applied".

https://github.com/DjangoGirls/djangogirls/issues/256


phalt:
On a 🎢 ! 👍

olasitarska:
Thanks!

PR #274, opened 2016-04-25
New read only fields in admin

patjouk:
I added a few fields in read only:

On event detail page:
* Is deleted: remove it to prevent organizers to accidentally delete their event.
* Is on homepage: they don't need to change it.

On event website page:
* Is deleted: remove it to prevent organizers to accidentally delete their website.
* Url: some organizers are changing their event website: it can create problem if they don't understand what they are doing. Sometime, they also chose an url that won't be compatible with a `clone-event` (ex: city2016).


olasitarska:
Awesome!

PR #277, opened 2016-05-13
adding workshop box to setup email

patjouk:
Adding info and link to the workshop box in the event setup email.


olasitarska:
@patjouk is it also worth adding discount to the shop here?

patjouk:
Done ;) (already configured on PythonAnywhere side)

PR #292, opened 2016-06-29
Adds unique_together for application form + email

### LINKED ISSUE #205: Prevent duplicate application entries

aniav:
These changes add uniqueness for email and form combination for applications. This fixes #205.
* a data migration to remove duplicates
* a migration adding the constraint
* form validation for the e-mail uniqueness


aniav:
@olasitarska @bmispelon I would like any of you to take a look at this ;)

applications/forms.py
```diff
@@ -16,22 +16,52 @@ def __init__(self, *args, **kwargs):
...
+        field_name = None
+
+        for name in self.cleaned_data:
+            question = None
+            pk = name.replace('question_', '')
+            value = self.cleaned_data[name]
+            try:
+                question = Question.objects.get(pk=pk, form=self.form)
+            except (Question.DoesNotExist, ValueError):
+                continue
```
olasitarska: This seems a little bit inefficient -- we're fetching a Question object for each field in the form...well, at least until we get to `email`... Maybe it'd be better to do `self.form.question_set.filter(question_type='email')` and compare just values of this field(s)?

aniav: Good point!

applications/migrations/0018_auto_20160628_2154.py
```diff
@@ -0,0 +1,30 @@
...
+    applications = Application.objects.values_list("form", "email").distinct()
+
+    for form, email in applications:
+        duplicates = (
+            Application.objects
+            .filter(form=form, email=email)
+            .values_list('id', flat=True)
+            .order_by('created')[1:]
+        )
+        Application.objects.filter(id__in=list(duplicates)).delete()
```
olasitarska: There it no much better way to do this, but I'm a little bit afraid we're gonna delete some accepted applicants. Maybe we should filter them out before deleting?

olasitarska: Oh this could be fixed by filtering by `('state', 'created')`, as accepted application will have state as `accepted`, which is first alphabetically from all available states :)

applications/forms.py
aniav: Additionally I think we should introduce some kind of checking if the form has just one e-mail field. Because currently if the form has two fields for e-mail (no idea what for, but events differ around the world) and the second one would be the applicant form... well, basically the wrong one would be attached to the `Application` object. But I have not idea how to change it at current stage so that it wouldn't take two days or something :D

PR #293, opened 2016-07-05
Creating secrets file

patjouk:
Since we can't use the WSGI file with management commands, I've created a `secrets.json` file to hide stuff :wink:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ static/build/
 static/local/
 static/rev-manifest.json
 temp/
+/djangogirls/secrets.json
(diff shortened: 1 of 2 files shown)


aniav:
Aren't we using the `postactivate` file when using management commands? The secrets are stored there and used e.g. with `deploy.sh`. I am not sure we need to add yet another file with secrets?

patjouk:
Ooooh, I didn't know that :heart_eyes_cat:.
I've tested your solution and it's working :smile:. I'll update the setting.py (shop code variable name is wrong) and clean wsgi files on PA.

aniav:
I am not sure how PA is working and if the WSGI runner will use the `postactivate`. Maybe that's the reason we have the variables in WSGI as well as postactivate? We will have to check :)

aniav:
✨

Issue #291, opened 2016-06-29
Default graphic for coach photo


helenst:
Some coaches don't supply a picture of themselves. If you upload no image for a coach, it doesn't line up very well on the site. It would be nice for it to display a default graphic if you don't upload anything. e.g. some kind of anonymous head outline, or perhaps something more creative.

aniav:
That is implemented now! :)

PR #296, opened 2016-07-21
Add empty coach image for #291

sodevious:
Adds a default image when there has been no coach image selected: https://github.com/DjangoGirls/djangogirls/issues/291


aniav:
Hey @sodevious the idea of using the heart as an empty coach photo is awesome! <3
I would like to suggest putting the photo link somewhere on the coach model, so when it is retrieved in other places it will still look the same. For example there is a coach list in the admin and we won't have the empty image there. Do you think you might make such change? ✨

sodevious:
Good idea! Let me try that :)

aniav:
@sodevious need any help with that? :)

aniav:
@sodevious I have used your changes and implemented the coach photo as a property on the model. Thank you for your help here! 👍 ❤️

PR #297, opened 2016-07-22
Fix #228 - Add organizers management views to admin

### LINKED ISSUE #228: Find a way to let organizer add co-organizers
The previous way to do it had a big problem: organizers could kick themselves out of their own event! Sadly, removing this right and ask them to contact the support team turned to be too time consuming.
Let's find a better way to do that! :wink:

olasitarska:
This pull request makes following changes:

- Adds a manage organizers view in admin, that allows superusers and organizers to remove organizers from upcoming events. 
- Adds an "add organizers" view in admin, that allows superuser and organizers to add organizers to upcoming events.
- Refactors "add_organizer" and "new_event" commands to use one form. Moves invitation to Slack into the form too.
- Adds sending automatic emails whenever organizer is added to event. This means we no longer need to send organizers passwords in emails manually. 
- Organizers can only add/remove organizers from their upcoming events. Superusers can do it for all upcoming events. 

@aniav your feedback on technical things here would be appreciated! I saw you recently refactored some of the commands too, so I wanna make sure we're consistent ✨ 
@patjouk your feedback on the functionality and whether it helps would be super helpful too!


olasitarska:
This is for issue #228

core/admin.py
```diff
@@ -1,16 +1,20 @@
...
+from datetime import datetime
+
+from django.contrib import admin, messages
 from django import forms
 from django.forms import ModelForm
 from django.contrib.auth import admin as auth_admin
 from django.contrib.flatpages.models import FlatPage
 from django.contrib.flatpages.admin import FlatPageAdmin, FlatpageForm
 from django.utils.safestring import mark_safe
+from django.http import Http404
```
aniav: Unused import 😜

core/admin.py
```diff
@@ -44,6 +48,84 @@ def get_readonly_fields(self, request, obj=None):
...
+        Custom admin view that allows user to remove organizers from an event
+        """
+        all_events = self._get_future_events_for_user(request)
+        event = self._get_event_from_get(request, all_events)
+
+        if 'remove' in request.GET and event in all_events:
+            user = User.objects.get(id=request.GET['remove'])
+            if user in event.team.all():
+                event.team.remove(user)
+                event.save()
```
aniav: You don't need to save the event object to have it's m2m relationships updated.

core/forms.py
```diff
@@ -14,6 +21,87 @@ def clean(self, values):
...
+    event = forms.ModelChoiceField(queryset=Event.objects.all())
+    name = forms.CharField(label="Organizer's full name")
+    email = forms.CharField(label="E-mail address")
+
+    def __init__(self, event_choices=None, *args, **kwargs):
+        super(AddOrganizerForm, self).__init__(*args, **kwargs)
+        if event_choices:
+            self.fields['event'].queryset = event_choices
+
+    def generate_password(self):
```
aniav: You can use [make_random_password](https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#django.contrib.auth.models.BaseUserManager.make_random_password) instead 🙂

core/forms.py
```diff
@@ -14,6 +21,87 @@ def clean(self, values):
...
+            user_invite(email, name)
+        except (ConnectionError, SlackerError) as e:
+            self._errors.append('Slack invite unsuccessful, reason: {}'.format(e))
+
+    def notify_existing_user(self, user):
+        content = render_to_string('emails/existing_user.html', {
+            'user': user,
+            'event': self.cleaned_data['event']
+        })
+        subject = 'You have been granted to new Django Girls event'
```
aniav: It should be "granted access" or "added" 😇

core/forms.py
```diff
@@ -14,6 +21,87 @@ def clean(self, values):
...
+        assert self.is_valid()
+        self._errors = []
+        email = self.cleaned_data['email']
+        event = self.cleaned_data['event']
+        if not User.objects.filter(email=email).exists():
+            self._password = self.generate_password()
+            first_name = self.cleaned_data['name'].split(' ')[0]
+            last_name = self.cleaned_data['name'].replace(first_name, '')
+            user = User.objects.create(email=email,
+                                        first_name=first_name,
```
aniav: Nitpick: the indentation doesn't match the previous line 😇

core/forms.py
```diff
@@ -14,6 +21,87 @@ def clean(self, values):
...
+                                        last_name=last_name,
+                                        is_active=True,
+                                        is_staff=True)
+            user.set_password(self._password)
+            user.save()
+            user.groups.add(1)
+            event.team.add(user)
+            event.save()
```
aniav: Sam as earlier, no need to save the event object :)

core/forms.py
```diff
@@ -14,6 +21,87 @@ def clean(self, values):
...
+    def send_email(self, content, subject, user):
+        send_mail(subject, content, "Django Girls <hello@djangogirls.org>",
+            [user.email], fail_silently=True)
+
+    def save(self, *args, **kwargs):
+        assert self.is_valid()
+        self._errors = []
+        email = self.cleaned_data['email']
+        event = self.cleaned_data['event']
+        if not User.objects.filter(email=email).exists():
```
aniav: Ideally you could use `User.objects.get_or_create` to avoid multiple queries to the User table 😀

templates/emails/setup.txt
```diff
@@ -21,10 +21,8 @@ Or you can forward incoming mail to your private email account by following thes
 
 Your website address is http://djangogirls.org/{{ event.eventpage.url }}
 
-Login here:
-http://djangogirls.org/admin/
-{% for member in members %}{{ member.first_name }} - email: {{ member.email }} password: {{ member.password }}
-{% endfor %}
+You should just receive a separate email with auto generated password and instructions to login.
```
aniav: Should be: "You should receive (...) in a moment" or "We have just sent you an e-mail" or "You should have just received" :)

olasitarska:
Oh I totally missed these comments! Will fix it all today 🔨

core/forms.py
```diff
@@ -14,6 +18,78 @@ def clean(self, values):
...
+
+    If user of given email already exists, they're added to the event and
+    receive e-mail notification about it.
+
+    If user is new, they're created (randomly generated password), invited
+    to Slack and receive e-mail notification with instructions to login
+    (including password).
+    """
+    event = forms.ModelChoiceField(queryset=Event.objects.all())
+    name = forms.CharField(label="Organizer's full name")
```
patjouk: I would change it for "Organizer's first and last name". That way, we will have the full name (some organizers only send the first name when I ask for full name) and it will be in the order we need.

core/admin.py
```diff
@@ -44,6 +49,83 @@ def get_readonly_fields(self, request, obj=None):
...
+        all_events = self._get_future_events_for_user(request)
+        event = self._get_event_from_get(request, all_events)
+
+        if request.method == 'POST':
+            form = AddOrganizerForm(all_events, request.POST)
+            if form.is_valid():
+                user = form.save()
+                messages.success(request,
+                    '{} has been added to your event, yay!'.format(user.get_full_name()))
+                return redirect('/admin/core/event/manage_organizers/?event_id={}'.format(event.id))
```
patjouk: It may be less confusing to return to the "add organizers page". They usually ask me to add a bunch of people at the same time.
There is also a problem with the redirection: if you have more than one event, it will redirect you to the first event in alphabetical order.

templates/admin/core/event/view_manage_organizers.html
```diff
@@ -0,0 +1,38 @@
+{% extends "admin/base_site.html" %}
+
+{% block content %}
+<h1 style="margin: 0 0 10px 0">Manage event organizers</h1>
```
patjouk: I would replace "manage" by "remove" since it's the only action you can do.

templates/emails/new_user.html
```diff
@@ -0,0 +1,25 @@
...
+
+<p>
+  Email: {{ user.email }}<br />
+  Password: {{ password }}
+</p>
+
+<p>Here you can find instructions on editing website: http://organize.djangogirls.org/website/</p>
+
+<p>We also invited you to Django Girls Slack, where we chat, communicate and meet each other!</p>
+
```
patjouk: This is absolutely awesome :heart_eyes: 
When I add someone to an existing event, I usually add info about the event email account: instructions on how to connect and that they should ask the main organizer to obtain the password.
It would be also cool to add the procedure for support ("You can contact hello@ if you need assistance but it’s usually faster to check our FAQ (link) or to post questions to our Google Group of Django Girls Organizers (link) or Slack channel (link) - there are more people that can help you! :)")

patjouk:
This PR is awesome :heart: It will definitely make my job easier.
I have three more comments I couldn't inline:
- For the new_event command: I moved the setup template email to a canned answer in gmail: it would be great if it only shows the thing I need to copy past (email recipients, email subject, event's emails and website). Since I'm not the only one to deploy event, I would totally understand if you want to keep the full text.
- A confirmation before removing someone from the team would be great: we removed this feature because of the miss-clicks and I'm pretty sure it will happen again. How about a page saying "are you sure you want to remove X from your team" or a checkbox with a save button?
- You shouldn't be able to kick yourself from a team :wink:

core/admin.py
```diff
@@ -44,6 +49,83 @@ def get_readonly_fields(self, request, obj=None):
...
+        """
+        all_events = self._get_future_events_for_user(request)
+        event = self._get_event_from_get(request, all_events)
+
+        if request.method == 'POST':
+            form = AddOrganizerForm(all_events, request.POST)
+            if form.is_valid():
+                user = form.save()
+                messages.success(request,
+                    '{} has been added to your event, yay!'.format(user.get_full_name()))
```
patjouk: It should also say that the new organizer received an email with their credentials and was also added to Slack.

templates/emails/new_user.html
olasitarska: Is there a standard response template that I can copy? Or maybe you can update those templates here and push to this branch?

olasitarska:
@patjouk From your non-inline comment:

> - For the new_event command: I moved the setup template email to a canned answer in gmail: it would be great if it only shows the thing I need to copy past (email recipients, email subject, event's emails and website). Since I'm not the only one to deploy event, I would totally understand if you want to keep the full text.

Done!

> - A confirmation before removing someone from the team would be great: we removed this feature because of the miss-clicks and I'm pretty sure it will happen again. How about a page saying "are you sure you want to remove X from your team", a checkbox with a save button or an "undo" button?

Done as well :)

> - You shouldn't be able to kick yourself from a team :wink: 

This too!

> - Organizers added by other organizers are not added to Google Group. Is there a way to obtain a list every day or week of the email addresses that were added?

I added `date_joined` to be displayed in [this table](https://djangogirls.org/admin/core/user/), so you're able to sort by it and see who recently joined. Also added this to filtering, so you can filter on who joined last week, for example ;) Hope that's enough!

PR #298, opened 2016-07-22
Fix #175 - Display search box instead of dropdown to add sponsor/coaches to EventPageContent

### LINKED ISSUE #175: Add sponsor functionality doesn't work
Hello!

1. G to add sponsor page in admin interface - https://djangogirls.org/admin/core/sponsor/add/
2. Fill all fields and click "save"
3. Then you will be redirected to sponsors list and you can see message about "new sponsor was added", but you will not see this sponsor in sponsors list, and else will not see this sponsor in event page.

I think that something broken with assigning sponsor to particular event.

Would be nice to have this fixed ASAP, because I have 2 sponsors who is waiting to be added to site.

Thanks

olasitarska:
I think this is gonna fix the problem described in #175.

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -112,12 +112,14 @@ class SponsorInline(admin.TabularInline):
     model = EventPageContent.sponsors.through
     extra = 1
     verbose_name_plural = 'Sponsors'
+    raw_id_fields = ["sponsor"]
 
 
 class CoachInline(admin.TabularInline):
     model = EventPageContent.coaches.through
     extra = 1
     verbose_name_plural = 'Coaches'
+    raw_id_fields = ["coach"]
 
 
 class EventPageContentAdmin(SortableModelAdmin):


aniav:
Hmm, I have just tested this and I am not sure this should look that way:
change_website_content___django_girls

I mean technically this is proper, but the sponsor/coach selection opens in a new window, I was not sure what to do to actually select the sponsor and back on the eventPageContent page it showed up as na ID, without the name, so adding mutliple sponsors you don't know what you have already added. The sponsors name would show up only after I saved the page. I am afraid this would bring more confusion than what we have now 🙁

aniav:
I have tried a few libraries for autocomplete for the sponsors and coaches but
1) none of them seemed to work properly in our setup
2) I feel this way it is implemented currently just isn't the right way to do that and should be fixed more broadly. What I mean by that is we should be using the meny-to-many itself and an autocomplete, we should not use the through model.

Other thing that should be done is sponsors merging in a data migration, so that their names don't repeat, which I would like to do :)

olasitarska:
Make sense! I'll close it for now then :)

PR #299, opened 2016-08-05
Added help text about pictures' licenses

patjouk:
I've added a reminder that cover pictures for events should always have a creative common license. Let's see if it helps :wink:


olasitarska:
Yay!

PR #302, opened 2016-08-30
Bugfix: email is an attribute on User object.

asendecka:
* we treated member (which was a User) as a dictionary, instead of object.


phalt:
👍

aniav:
🎉

PR #303, opened 2016-08-31
Add a 'short email' option for new_event command

patjouk:
I've added an option `--short` or `-s` to the `new_event` command: by default, the whole setup email will be displayed. If you use this option, you'll only get event's email and website.

### DIFF
--- a/core/management/commands/new_event.py
+++ b/core/management/commands/new_event.py
@@ -152,7 +152,8 @@ def brag_on_slack_bang(city, country, team):
 
 
 @click.command()
-def command():
+@click.option('--short', '-s', is_flag=True, help="Shorter version of the setup email to use with a canned email.")
+def command(short):
     """Creates new Django Girls event"""
     # Basics
     (city, country, date, url, event_mail) = get_basic_info()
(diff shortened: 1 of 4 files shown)


olasitarska:
@patjouk looks good! could you add a test for this here as well? https://github.com/DjangoGirls/djangogirls/blob/master/core/tests/test_commands.py#L16

patjouk:
So, trying to write test for this was fun but I run into a problem:

The new_event command seems to always [be failing on Travis](https://codecov.io/gh/DjangoGirls/djangogirls/src/8da41d3303d4bde4488f22129e1cd3905b2d36f5/core/management/commands/new_event.py#L184). This is linked to a call to `core.default_eventpage_content.get_random_photo`, that tries to get a random photo from `static/build`, which is created by gulp (not run on Travis), and fails.

I'm not sure what is the best way to fix that :disappointed: 
Thanks @madjar for the awesome help.

aniav:
@patjouk we should probably add the step of running gulp. Do you want me to help here? :)

patjouk:
@aniav I would love your help :wink: Thanks!

olasitarska:
Ok, so I think to add gulp to Travis, we need to:

- Add `npm install` as another line in `install` section of [Travis config](https://github.com/DjangoGirls/djangogirls/blob/master/.travis.yml#L12)
- Add `gulp build` as another line in `before_script` section of [Travis config](https://github.com/DjangoGirls/djangogirls/blob/master/.travis.yml#L18)

And cross our fingers that this will work? ;)

olasitarska:
Closing in favour of #303. I fixed this issue by pointing Travis on source static files instead of built ones.

PR #305, opened 2016-09-05
Redo of the workshop box page

patjouk:
Since the workshop box is now on the store, I've changed and adapted the previous text.
I also added a healthy dose of emoji, because why not :wink:


olasitarska:
Thanks! <3

PR #308, opened 2016-09-15
Creating a page for GitHub sponsoring requests

patjouk:
Since my instructions for asking a GitHub grant were too confusing (sorry about that :sweat:), I've created a new page with new instructions and an embedded Typeform. You can access this page only if you're  logged in. It also shows the next deadline to let organizers know when their requests will be sent.


templates/event/sponsor-request.html
```diff
@@ -0,0 +1,35 @@
...
+                        <a href="https://djangogirls.org/events/">check if an event already happened in your city</a>
+                        and if they received money from GitHub (check their sponsors section).
+
+                    <p>If your event is the first in your city or if the previous ones didn't receive sponsoring
+                    from GitHub, you can apply.</p>
+
+                    <p>Note: Filling this form will be long. You'll have to download and edit some documents (invoice,
+                        bank information, etc).</p>
+
+                    <p>We send those requests bi-weekly to GitHub. <b>Next deadline: {{ deadline }}.</b></p>
```
bmispelon: I'd use "once every two weeks" instead of "bi-weekly" to help with the dual meaning of that word

core/utils.py
```diff
@@ -49,3 +51,12 @@ def get_approximate_date(date_str):
         except ValueError:
             return None
     return None
+
+def next_deadline():
+    """Return the next deadline for GitHub sponsoring so the organizers know when we will send their request."""
+    now = datetime.now()
```
bmispelon: Because we're only working with dates (not datetimes), you can use `today = date.now()` (`from datetime import date`).

core/utils.py
```diff
@@ -49,3 +51,12 @@ def get_approximate_date(date_str):
...
+    start_date = datetime.strptime('04/09/16', '%d/%m/%y')
```
bmispelon: You can build a `date` object directly with `start_date = date(2016, 9, 4)`

core/utils.py
```diff
@@ -49,3 +51,12 @@ def get_approximate_date(date_str):
...
+
+    all_deadlines = (start_date + timedelta(days=14 * i) for i in itertools.count())
```
bmispelon: This is really clever!

templates/event/sponsor-request.html
```diff
@@ -0,0 +1,35 @@
+{% extends 'global/base.html' %}
+{% load staticfiles %}
+
+{% block title %}Request for GitHub sponsoring{% endblock %}
+{% block description %}Fill the form available on this page to apply for a GitHub grant.{% endblock %}
+
+{% block content %}
```
bmispelon: Can you fix the indentation of that block? It seems to be indented by one level too much.

core/utils.py
bmispelon: And you can do the same thing in the tests too: you don't need to use `strptime` to create the date objects.

templates/not_logged.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'global/base.html' %}
+
+{% block content %}
+    <div class="container">
+        <div class="row">
+            <div class="col-md-12">
+                <h2>This page is restricted to Django Girls workshops' organizers</h2>
+
+                <p>If you're an organizer, you need to <a href="https://djangogirls.org/admin">log in</a> to see this page.</p>
```
bmispelon: If we manage to use `@login_required` then this template should not be necessary. But if that doesn't work, you can use `{% url 'admin:index' %}` instead of hardcoding the link.

bmispelon:
Here's an implementation of `next_deadline` that doesn't rely on an arbitrarily long generator (`itertools.count`):
```
def next_sunday(day):
    """
    Return a date object corresponding to the next Sunday after the given date.
    If the given date is a Sunday, return the Sunday next week.
    """
    if day.weekday() == 6:  # sunday
        return day + timedelta(days=7)
    else:
        return day + timedelta(days=(6 - day.weekday()))


def next_deadline():
    """
    Return the next deadline when we need to send invoices to GitHub.
    Deadlines are every second Sunday, starting from September 4th 2016.
    """

    today = date.today()

    days_since_starting_sunday = (today - date(2016, 9, 4)).days

    if days_since_starting_sunday % 14 < 7:
        return next_sunday(next_sunday(today))
    else:
        return next_sunday(today)
```

patjouk:
PR is ready for a new round of review :wink: 
Thanks @bmispelon for the help and comments.

templates/event/sponsor-request.html
```diff
@@ -0,0 +1,33 @@
...
+    <section id="flatpage">
+        <div class="row clearfix">
+            <div class="col-md-8 col-md-offset-2 col-sm-10 col-sm-offset-1 col-xs-10 col-xs-offset-1">
+                <h1>Request for GitHub sponsoring</h1>
+
+                <p>GitHub can sponsor one Django Girls workshop per city with a 500$ grant.</p>
+
+                <p>Before applying, your website needs to be ready and online. Then, you need to
+                    <a href="https://djangogirls.org/events/">check if an event already happened in your city</a>
```
bmispelon: Can you use the `{% url ... %}` tag here instead of `https://djangogirls.org/events/`. Look inside the `urls.py` file for the name of the URL to use.

bmispelon:
I left one final comment about using `{% url ... %}`. Once that's fixed, you have my :+1: to merge.

Thanks! :sparkles:

templates/event/sponsor-request.html
patjouk: oops, forgot to fix it ><'

bmispelon:
:tada:

PR #312, opened 2016-11-05
Fixes #309 "Access to Django Girls website" HTML email sent as plain text

### LINKED ISSUE #309: "Access to Django Girls website" HTML email sent as plain text
I've received the "Access to Django Girls website" from hello@djangogirls.org to get the access details of the new event's website.

This email is HTML, but it is displayed as plain text:

``` html
<p>Hi there!</p>

<p>Congrats! You're now a Django Girls organizer, so awesome! Here is access to your
website:</p>

[...]
```

Looking at the email's body, it is sent with `Content-Type: text/plain; charset=utf-8` which I think would be the cause of the issue.

Looking at the site's codebase, I think this is where the problem is: https://github.com/DjangoGirls/djangogirls/blob/8da41d3303d4bde4488f22129e1cd3905b2d36f5/core/forms.py#L66 

``` python
    def notify_new_user(self, user):
        content = render_to_string('emails/new_user.html', {
            'user': user,
            'event': self.cleaned_data['event'],
            'password': self._password,
            'errors': self._errors,
        })
        subject = 'Access to Django Girls website'
        self.send_email(content, subject, user)

    def send_email(self, content, subject, user):
        send_mail(subject, content, "Django Girls <hello@djangogirls.org>",
                  [user.email])
```

The email content is rendered from HTML templates, but is sent with `send_mail`'s `message` instead of `html_message`. See https://docs.djangoproject.com/en/1.10/topics/email/#send-mail.

EmadMokhtar:
Change send_mail() method and add 2 tests for it.


olasitarska:
Thank you @EmadMokhtar! :)

PR #313, opened 2016-11-05
added option to exclude existing recipients from receiving emails

burhan:


olasitarska:
Hi @burhan! Thanks for this PR. I think for this to actually work we need to enable sending the same email multiple times, as currently when email is sent, you cannot click "send" on it again. I tried testing this solution locally and couldn't use your feature :(

olasitarska:
Closing this due to inactivity -- please let me know if you want to work on this again and we can reopen :)

PR #314, opened 2016-11-05
Menu that works on mobile devices

ludw:
I replaced the current navigation on the main webpage with one that works better on mobile devices.

The actual hamburger menu I used is the same one used for the event pages. 

As far as I can see the styling looks identical on desktop.


patjouk:
Thanks! :sparkles:

PR #315, opened 2016-11-06
Apply isort with standard configurations

EmadMokhtar:
I ran isort on the project and there are many changes need to be done.

Note:

I used the standard isort configurations


olasitarska:
Thank you!

Issue #306, opened 2016-09-06
Broken layout on mobile


wattawak:
on https://djangogirls.org/denver/, there is a horizontal scroll and whitespace on the right-hand side of the screen when viewing the site on mobile

wattawak:
@ludw you can see that we did set max-width:100% on the iframe for the map, but we're still having the issue

ludw:
On what devices and browsers do you have this issue?

wattawak:
iPhone 6s // Safari

ludw:
Okay, I did some investigation and it looks like it's the word "programming" in the H1 at the top that get's to big. 

I think the best way to solve this is to add some css to the event page to make the text smaller on mobile devices. I'll see if I can get a patch for that merged since this seems to be an issue for pretty much all event pages. 

If you want a quick workaround you could try setting overflow-x: hidden on the surrounding div or something like that.

PR #317, opened 2016-11-10
Changed size of h1 and h2 in #about to not break on iPhone 6S

ludw:
I changed the media queries for the headings in the about box so the heading gets smaller earlier.

The word "programming", which is used by default in the H1, was wide enough to break the page on some mobile devices with the previous css (Issue #306).


patjouk:
Thanks! :sparkles:

PR #319, opened 2016-11-20
Add menu item to previous event in the same city

### LINKED ISSUE #276: Links between previous and new events
When a city has more than one event, it would be cool to have a link to the previous event in the menu.

We could do it manually but it would be better to update [this script](https://github.com/DjangoGirls/djangogirls/blob/master/core/management/commands/copy_event.py).

joncasdam:
When copying an event via `./manage.py copy_event`, adds a menu item for the last event in the same city.

Fix #276

### DIFF
--- a/core/management/commands/copy_event.py
+++ b/core/management/commands/copy_event.py
@@ -45,6 +45,7 @@ def command():
     # Gather data
     (event, number, date) = gather_information()
     eventpage = event.eventpage
+    previous_url = "/{}".format(eventpage.url)
     organizers = event.team.all()
 
     # Print stuff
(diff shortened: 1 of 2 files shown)


aniav:
Hi @joncasdam. Sorry you had to wait so long for the review of your PR. There are a few problems with the idea of adding previous event links to the menu:
- it won't work properly if the organizers of the second event will change the order of the elements in the menu list.

We could try to overcome this issue by search by the menu item title, but we would still fail with this when the event would have the element translated by the organizers (which happens quite often).

Additionally I think we have come to the point where we need to store the event number in an additional field and generate links basing on that number. We might also put the link in the newly created fixed footer.

Keeping all this in mind I can't merge your changes. Thank you for your work and sorry it took so long! 🤗

joncasdam:
hello, @aniav. No problem. So the related issue should be closed too?

aniav:
@joncasdam I think the issue should stay, as we still would like to have the link to the previous event, just not as something generated for each event when it's being copied :)

PR #320, opened 2016-11-21
Moving the 'how to organize' section up

patjouk:
"Don’t see an event in your city?" section is at the bottom of the page: the events list got bigger (:tada:) so it might be a good idea to bring it up between the next and past events.


olasitarska:
Yay! 🌈

PR #321, opened 2016-11-22
small fixes in event page templates

patjouk:
- Removed the coach form: some events don't change it and keep this old link.
- "Django Girls Australia..." to "This workshop": some events don't change it.
- Fix typo in footer


olasitarska:
Nice one! Thank you Lucie :)

PR #322, opened 2016-11-25
Fixed #287 made it possible to change the email address of applicants…

### LINKED ISSUE #287: make it possible to change the email of applicants
We have a few applications with typos in the email field but we can't change them.

elifkus:
… through the admin interface

Followed @bmispelon 's advice (instead of removing the fields from the form I made the fields optional)


elifkus:
fixes #287

patjouk:
Thanks :sparkles:

PR #326, opened 2016-12-02
Fix toggling of the last question in the FAQ

rhcarvalho:
The div id was duplicated from the first question, so that clicking on the last question was toggling the answer for the first, and there was no way to see the last answer.


helenst:
Thank you! :rose:

PR #327, opened 2016-12-04
Answer model defined as inline in admin

beratdogan:
This PR just adding an inline to see answers of an application in very compact form.

screen shot 2016-12-04 at 15 12 47


olasitarska:
Thank you, quite useful! :)

PR #328, opened 2016-12-04
Optimized applications page

beratdogan:
This PR includes some performance improvements on applications page.

Basicly I did made following two changes:
- Prefetch scores in bulk to prevent fetching all scores for every application
- Changed `is_scored_by_user` to detect and use score data already we fetched

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -232,7 +232,11 @@ class Application(models.Model):
         Returns true if the given user has scored this application
         or false if they have not, or there is a zero score.
         """
-        return self.scores.filter(user=user, score__gt=0).exists()
+        try:
+            self._prefetched_objects_cache['scores']  # check if scores prefetched
+            return any(map(lambda s: s.user_id == user.id, self.scores.all()))
+        except (AttributeError, KeyError):
+            return self.scores.filter(user=user, score__gt=0).exists()
 
     def __str__(self):
         return str(self.pk)
--- a/applications/models.py
+++ b/applications/models.py
@@ -227,15 +227,15 @@ class Application(models.Model):
     def is_accepted(self):
         return self.state == 'accepted'
 
-    def is_scored_by_user(self, user):
+    def is_scored_by_user(self, user, use_prefetched=True):
         """
         Returns true if the given user has scored this application
         or false if they have not, or there is a zero score.
         """
         try:
-            self._prefetched_objects_cache['scores']  # check if scores prefetched
+            assert self._prefetched_objects_cache['scores'] and use_prefetched # check if scores usable
             return any(map(lambda s: s.user_id == user.id and s.score and s.score > 0, self.scores.all()))
-        except (AttributeError, KeyError):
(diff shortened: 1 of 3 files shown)


applications/models.py
```diff
@@ -232,7 +232,11 @@ def is_scored_by_user(self, user):
         Returns true if the given user has scored this application
         or false if they have not, or there is a zero score.
         """
-        return self.scores.filter(user=user, score__gt=0).exists()
+        try:
+            self._prefetched_objects_cache['scores']  # check if scores prefetched
+            return any(map(lambda s: s.user_id == user.id and s.score and s.score > 0, self.scores.all()))
+        except (AttributeError, KeyError):
```
helenst: Thank you, it would be great to reduce the number of queries here.

I think `_prefetched_objects_cache` is an undocumented Django internal. To ensure we pick up on any changes that may happen in future versions of Django, and also to verify the logic here, this branch of the code should be covered by a test. Could you add one please?

The lambda would probably also be clearer as a named function.

Thanks! :sparkles:

beratdogan:
@helenst I updated and added some tests. Tests cases enough?

I didn't move lambda to a named function because it's very inner logic. I have no name idea for it.

applications/models.py
```diff
@@ -227,12 +227,16 @@ def get_by_rsvp_code(self, code, page):
...
+    def is_scored_by_user(self, user, use_prefetched=True):
         """
         Returns true if the given user has scored this application
         or false if they have not, or there is a zero score.
         """
-        return self.scores.filter(user=user, score__gt=0).exists()
+        try:
+            assert self._prefetched_objects_cache['scores'] and use_prefetched # check if scores usable
+            return any(map(lambda s: s.user_id == user.id and s.score and s.score > 0, self.scores.all()))
+        except (AssertionError, AttributeError, KeyError):
```
helenst: This is a lot of different error types to be catching, I don't feel like it's clear what's going on and what the code's expecting to fail. I can see that you're catching AttributeError in case `prefetched_objects_cache` isn't present, but it'll also trigger on other things - it's easy to obscure legitimate errors this way when you probably wanted to hear about them.

How about an alternative way of checking the cache, which doesn't require try/except

```python
cache = getattr(self, '_prefetched_objects_cache', {})
if 'scores' in cache:
    # etc...
```

aniav:
Hmm, I thought the ✔️ would disappear on the PR list if I request changes. So I actually added myself here without reason. Sorry for that :)

aniav:
Hi @beratdogan the idea of optimising the applications page is great but I believe the main problem with the amount of queries there is that `scored_by_user` templatetag is producing queries. Your idea is to fix this problem by using internal cache of querysets. I think this is not the best approach. So what I am going to do is 1) close this PR 2) push some `prefetch_related` and `select_related` optimisations to the website 3) create an issue to change the code checking if a person has scored an application to not use database queries in templatetags.

I imagine the properly working code should query the database first and anonymise the data on the python side, without any further queries.

PR #330, opened 2016-12-05
Define 'city' at top of decorator

ulgens:
`city` value used several times in related decorator. Defining it at top is better for both readability and preventing duplication.


helenst:
This looks great, thank you! :panda_face:

PR #335, opened 2017-01-06
Fixed incorrect pluralization of model name in admin

bmispelon:


core/models.py
```diff
@@ -406,3 +406,7 @@ class Story(models.Model):
 
     def __str__(self):
         return self.name
+
+    class Meta:
+        verbose_name = "story"
+        verbose_name_plural = "stories"
```
patjouk: Are you sure it's not storys?

bmispelon:
Danke! 🐌

PR #338, opened 2017-01-07
Clean requirements, introduce requirements.in

patrys:
Together with [`pip-compile`](https://github.com/nvie/pip-tools) we can have reproducible builds and an easy upgrade mechanism.

This needs to be merged after #337 as it drops some of the requirements only used by jobs and meetups.

## Rationale

Pinning exact versions is important for repeatable builds that run the same code the developers and CI environments used to conduct testing.

At the same time it's important to know which of the dependencies belong to the project itself and which are mere implementation details of our direct dependencies, otherwise we get an ever-growing list of dependencies which soon turns into a pile of technical debt.

`pip-compile` allows us to cleanly separate those two while still being able to include information such as "it's known not to work with django >= 1.11" in `requirements.in`.

Working with `requirements.in` is easy as the list is concise and if in doubt it's usually enough to `git grep` the codebase and see if a particular dependency is safe to remove. `pip-compile` will do the rest and figure out whether the package can be safely uninstalled (as it could still be needed as an indirect dependency).


olasitarska:
This is cool! Can we merge this after the sprint though, so that everyone doesn't have to update their environments again today?

patrys:
Sure, tell me when and I'll be happy to rebase.

aniav:
I think we can merge it now. Can you rebase? :)

PR #341, opened 2017-01-07
Remove Postmortem model

oinopion:
This adds stats fields to Event and removes `Postmortem` model completely.


core/fixtures/core_views_testdata.json
```diff
@@ -78,7 +78,7 @@
         "city": "\u0141\u00f3d\u017a",
         "name": "Django Girls \u0141\u00f3d\u017a",
         "country": "Poland",
-        "latlng": "51.768749, 19.4569817",
```
oinopion: Looks like we:
1) Make HTTP requests during tests
2) Łódź moved around

oinopion:
<3

PR #344, opened 2017-01-07
Create default event data when a new event is created in the admin panel

patrys:
Moved the responsibility of new data creation to the form.
Made the management command and the ModelAdmin use the form.

### DIFF
--- a/core/admin.py
+++ b/core/admin.py
@@ -15,8 +15,8 @@ from django.utils.safestring import mark_safe
 from suit.admin import SortableModelAdmin
 
 from .filters import OpenRegistrationFilter
-from .forms import (AddOrganizerForm, UserChangeForm, UserCreationForm,
-                    UserLimitedChangeForm)
+from .forms import (AddOrganizerForm, EventForm, UserChangeForm,
+                    UserCreationForm, UserLimitedChangeForm)
 from .models import (Coach, Event, EventPageContent, EventPageMenu,
                      Postmortem, Sponsor, Story, User)
 
@@ -27,6 +27,7 @@ class EventAdmin(admin.ModelAdmin):
     list_filter = (OpenRegistrationFilter,)
     search_fields = ('city', 'country', 'name')
     filter_horizontal = ['team']
+    form = EventForm
 
     def get_queryset(self, request):
         qs = super(EventAdmin, self).get_queryset(request)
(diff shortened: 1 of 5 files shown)


core/management/commands/new_event.py
```diff
@@ -6,13 +6,11 @@
...
-                                            get_default_menu)
-from core.forms import AddOrganizerForm
-from core.models import Event, EventPageContent, EventPageMenu
-from core.slack_client import slack
-from core.utils import get_coordinates_for_city
+from ...command_helpers import gather_event_date_from_prompt
+from ...forms import AddOrganizerForm, EventForm
+from ...models import Event
+from ...slack_client import slack
+from ...utils import get_coordinates_for_city
```
olasitarska: I think this import style makes it less readable.

core/admin.py
```diff
@@ -199,6 +200,12 @@ def view_add_organizers(self, request):
             'form': form,
             'title': 'Add organizers',
         })
+    
+    def save_model(self, request, obj, form, change):
+        created = not obj.pk
+        super(EventAdmin, self).save_model(request, obj, form, change)
+        if created:
+            form.add_default_data(obj)
```
olasitarska: A little bit confused why we need this on both EventForm and EventAdmin, as EventAdmin is using EventForm anyway. There is probably a reason though, right?

core/management/commands/new_event.py
olasitarska: I'm not feeling too strongly about it though, so feel free to ignore that comment as well.

core/admin.py
patrys: `ModelAdmin` calls all form saves with `commit=False` as it expects the instance to be saved later using `save_model`.

olasitarska: ah, I see. make sense! :) could you put that in the comment there maybe?

core/management/commands/new_event.py
aniav: This is something I proposed. It's good because it makes the imports independent of the package name. If we decide to rename the `core` app it will help.

As for renaming the name of this app should be something more meaningful, like `events` and the User related models and views should be moved to some other app. Currently when people don't know where to add something (like view or a model) they add it to core and we end up with this huge app with basically everything. Moving models around between apps can be done with migrations and it's pretty straightforward, so I would be happy to do it in some point in the future :)

olasitarska: @aniav :+1: then! :)

olasitarska:
Thanks!

PR #347, opened 2017-01-07
Add prototype code for gmail account creation

helenst:
Needs to be integrated into something! Also relies on a credentials json file which must be separately supplied.

### DIFF
--- a/prototype/gmail_accounts.py
+++ b/prototype/gmail_accounts.py
@@ -1,0 +1,101 @@
+import string
+from httplib2 import Http
+from random import choice
+
+from apiclient.errors import HttpError
+from apiclient.discovery import build
+from oauth2client.service_account import ServiceAccountCredentials
+
+
+PASSWORD_CHARS = string.ascii_letters + string.digits
+
+def generate_password(length=10):
+    return ''.join([choice(PASSWORD_CHARS) for i in range(length)])
+
+
+
+"""
+Several things were needed to get this working:
+1. Create an app in Developer Console
+2. Create a service account to enable 2 legged oauth (https://developers.google.com/identity/protocols/OAuth2ServiceAccount)
+3. Enable delegation of domain-wide authority for the service account.
+4. Enable Admin SDK for the domain.
+5. Give the service account permission to access admin.directory.users service (https://admin.google.com/AdminHome?chromeless=1#OGX:ManageOauthClients).
+"""
+
+SCOPES = 'https://www.googleapis.com/auth/admin.directory.user'
+
+credentials = ServiceAccountCredentials.from_json_keyfile_name(
+    'djangogirls_website-93134565f52d.json',
+    scopes=SCOPES
+)
+
+delegated_credentials = credentials.create_delegated('hello@djangogirls.org')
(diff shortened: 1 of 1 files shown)


prototype/gmail_accounts.py
```diff
@@ -0,0 +1,101 @@
...
+from random import choice
+
+from apiclient.errors import HttpError
+from apiclient.discovery import build
+from oauth2client.service_account import ServiceAccountCredentials
+
+
+PASSWORD_CHARS = string.ascii_letters + string.digits
+
+def generate_password(length=10):
```
bmispelon: There's a `get_random_string` function in `django.utils.crypto`, maybe we can reuse that?

helenst: Did not know about that. Sounds like a good idea - thanks! :+1:

prototype/gmail_accounts.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # The old email address is kept as an alias to the new one, but we don't want this.
+    service.users().aliases().delete(userKey=new_email, alias=old_email).execute()
+
+    print('Account %s renamed to %s' % (old_email, new_email))
+
+    return response
+
+
+def get_account(slug):
```
bmispelon: What about calling this function `print_account` instead?

prototype/gmail_accounts.py
```diff
@@ -0,0 +1,101 @@
...
+2. Create a service account to enable 2 legged oauth (https://developers.google.com/identity/protocols/OAuth2ServiceAccount)
+3. Enable delegation of domain-wide authority for the service account.
+4. Enable Admin SDK for the domain.
+5. Give the service account permission to access admin.directory.users service (https://admin.google.com/AdminHome?chromeless=1#OGX:ManageOauthClients).
+"""
+
+SCOPES = 'https://www.googleapis.com/auth/admin.directory.user'
+
+credentials = ServiceAccountCredentials.from_json_keyfile_name(
+    'djangogirls_website-93134565f52d.json',
```
olasitarska: I understand that this file will need to be deployed to pythonanywhere?

helenst: Yes - I guess it will need to be put into place manually.

prototype/gmail_accounts.py
```diff
@@ -0,0 +1,101 @@
...
+
+def create_account(slug, city_name):
+    """
+    Create a new account
+
+    e.g. create_account('testcity', 'Test City')
+    """
+    email = make_email(slug)
+    password = generate_password()
+    response = service.users().insert(body={
```
aniav: What happens if a specified e-mail already exists? We should be able to catch the error and, for example, display it in the admin as a ValidationError

prototype/gmail_accounts.py
olasitarska: Make sense :)

prototype/gmail_accounts.py
helenst: It'll raise an HttpError (with code 409). I'll document that in the comment for now.

aniav: Ah, great! 💃

PR #351, opened 2017-01-07
Use VCR.py to freeze all external HTTP calls

### LINKED ISSUE #346: We should be mocking all requests to external APIs in tests
To reproduce: run tests without internet connection :-)

patrys:
Fixes #346

### DIFF
--- a/applications/tests/test_forms.py
+++ b/applications/tests/test_forms.py
@@ -1,5 +1,6 @@
 import os
 
+import vcr
 from django.test import TestCase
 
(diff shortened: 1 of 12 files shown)


helenst:
Merging this because it's awesome. Thank you! :1st_place_medal: 

However, there's still an HTTP request that happens during tests - it's in test_adding_organizer_as_superuser core/tests/test_admin.py and it's trying to do stuff on slack - so will keep the issue open.

PR #352, opened 2017-01-07
Reminder and thankyou emails

kirberich:
There's a lot of logic in the filters for deciding when emails should be sent for which events. If you can think of any tests to add let me know!


kirberich:
One caveat: events from the past won't get reminders to submit data because I didn't want to make the migration even more difficult - so there might be a few cases of events from the last few months where manual prodding is required.

Otherwise the migrations seem to make sense now, but when migrating this best do a manage.py handle_emails manually to check which events would get emails when deployed.

core/migrations/0030_email_flags.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0029_auto_20170107_1539'),
+    ]
+
+    operations = [
+        migrations.RunPython(set_default_email_flags),
```
bmispelon: It'd be nice to add a reverse operation (even if it's `migrations.RunPython.noop`) so we can run this migration backwards if needed.

See https://docs.djangoproject.com/en/1.10/ref/migration-operations/#django.db.migrations.operations.RunPython.noop

core/management/commands/handle_emails.py
```diff
@@ -0,0 +1,128 @@
...
+)
+from django.template.loader import get_template
+from django.utils import timezone
+
+from core.models import Event
+
+
+def send_event_emails(events, subject, plain_template, html_template, timestamp_field, email_type,
+                      ignore_approximate_events=False):
+    """Send out any thank you emails that need sending."""
```
olasitarska: Comment needs updating as this is a generic method now

core/migrations/0030_email_flags.py
olasitarska: Can't see how not having the reverse migration is gonna hurt us, so unless you feel strongly about this, I'd just merge it this without that.

PR #354, opened 2017-01-07
[WIP] Added page for auto-inviting translators to slack

bmispelon:
This still needs some work:

* [ ] Some tests (adding tests for the slack API might prove tricky but we should still be able to have some unit tests in a few places
* [ ] Better copy for the form and the success page
* [ ] A nicer looking form


aniav:
I'll close this one, doesn't seem anyone is doing anything with it 🙌

PR #355, opened 2017-01-08
[WIP] Triaging new city requests

asendecka:


core/models.py
```diff
@@ -61,6 +63,18 @@ class Meta:
...
+    def invite_to_slack(self):
+        user_invite(self.email, self.first_name)
+
+    def generate_password(self):
+        password = User.objects.make_random_password()
+        self.set_password(password)
+        return password
+
+    def add_to_organizers_group(self):
+        group = Group.objects.get(name="Organizers")
```
olasitarska: This is gonna fail if organizers group doesn't exist

core/models.py
```diff
@@ -192,6 +213,78 @@ def delete(self):
         self.is_deleted = True
         self.save()
 
+    def add_default_content(self):
```
aniav: I think this should be separated from the model. We should run it once per every new instance, keeping it in the model might end up calling it by mistake. Keeping it as a function separated from the event would be enough, I think :)

organize/models.py
```diff
@@ -28,20 +35,144 @@ class EventApplication(models.Model):
...
     class Meta:
         permissions = (
             ("can_accept_organize_application",
              "Can accept Organize Applications"),
         )
 
+    def __str__(self):
+        return "{}, {} ({})".format(self.city, self.country, self.get_status_display())
+
+    def create_event(self):
```
aniav: Does this belong to the EventApplication model? This has nothing to do with the EventApplication itself, should be in the core app, I think :)

organize/models.py
```diff
@@ -28,20 +35,144 @@ class EventApplication(models.Model):
...
+        """
+        Rejecting event in triaging. Performs following actions:
+        - changes status to REJECTED
+        - sends a rejection email
+        """
+        if not self.status == REJECTED:
+            self.change_status_to(REJECTED)
+            self.send_rejection_email()
+
+    def send_rejection_email(self):
```
aniav: Maybe we could move the e-mails to emails.py inside this app?

organize/migrations/0004_remove_eventapplication_rejection_reason.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
aniav: Let's squash the migrations into one before merging :)

organize/admin.py
```diff
@@ -1,14 +1,155 @@
...
+            obj.main_organizer_last_name,
+            obj.main_organizer_email
+        )
+
+    def get_urls(self):
+        urls = super(EventApplicationAdmin, self).get_urls()
+        my_urls = [
+            url(r'(?P<application_id>\d+)/triage/(?P<new_status>[\w\d/]+)/$',
+                self.admin_site.admin_view(self.view_change_application_status),
+                name='organize_eventapplication_change_application_status'),
```
aniav: `organize` is a napespace here, you can cut it out from the name :)

core/forms.py
```diff
@@ -43,61 +37,14 @@ def __init__(self, event_choices=None, *args, **kwargs):
...
-            user.is_active = True
-            user.set_password(self._password)
-            user.save()
-            user.groups.add(1)
-            self.invite_to_slack(email, user.first_name)
-            self.notify_new_user(user)
-        else:
-            self.notify_existing_user(user)
+        first_name = self.cleaned_data['name'].split(' ')[0]
+        last_name = self.cleaned_data['name'].replace(first_name, '')
```
bmispelon: This will give strange results for a name like `"Jo Johnson"`. Something like `first_name, _, last_name = self.cleaned_data['name'].partition(' ')` should work better and handle edge cases too (no space present in the name, more than one space present, ...)

core/models.py
```diff
@@ -192,6 +213,78 @@ def delete(self):
...
+        )
+        password = None
+        if created:
+            password = user.generate_password()
+            user.add_to_organizers_group()
+
+        self.invite_organizer_to_team(user, created, password)
+        return user
+
+    def create_organizers(self, team, email_password):
```
aniav: Add organizer totally belongs here as it can be reaused on an instance `create_organizers` might find a better place.

I think we could just create a file called something like initial_event_data and put the functions there. They might be called just like from here. I somehow feel that putting non-reusable methods later available on the model instance is not the best idea :)

organize/admin.py
```diff
@@ -1,14 +1,155 @@
...
+    def view_change_application_status(self, request, application_id, new_status):
+        """
+        Custom EventApplication admin view for handling triaging
+        """
+        application = get_object_or_404(EventApplication, id=application_id)
+
+        if new_status in [IN_REVIEW, ON_HOLD]:
+            application.status = new_status
+            application.status_changed_at = timezone.now()
+            application.save()
```
bmispelon: Using `application.save(update_fields=['status', 'status_changed_at'])` should help with race conditions.

organize/models.py
```diff
@@ -28,20 +35,144 @@ class EventApplication(models.Model):
...
+        """
+        emails = [coorganizer.email for coorganizer in self.coorganizers.all()]
+        emails.append(self.main_organizer_email)
+        return emails
+
+    def change_status_to(self, status):
+        """ Changes status to the status provided
+            - sets proper status_changed_at datetime
+        """
+        self.status = REJECTED
```
olasitarska: Status should come from `status` not hardcoded to `rejected`

organize/admin.py
olasitarska: I don't think so -- namespace is `admin` in this case

organize/admin.py
olasitarska: Actually, I need to reuse `Event.change_status_to` in this case and add the partial update there.

organize/admin.py
aniav: Ah, right! Good catch :)

organize/models.py
olasitarska: I feel like if we're creating Event out of EventApplication, it may as well be here... but not feeling strongly :)

organize/models.py
olasitarska: Yup, will do it :)

organize/models.py
```diff
@@ -28,20 +35,144 @@ class EventApplication(models.Model):
...
+        self.status_changed_at = timezone.now()
+        self.save()
```
bmispelon: Using `self.save(update_fields=['status', 'status_changed_at'])` here would be a bit nicer but it's not a strictly equivalent change. I don't know if some parts of the code rely on all fields (not just status) to be saved.

organize/admin.py
olasitarska: ✨

organize/models.py
aniav: Methods on the model instance should basically modify it's state somehow or help retrieving it's values. Other things might as well be moved elsewhere. I feel all the methods related to creating the fresh event instance should just be put in a simple python file in one place in the core app. Now we have create_event here, add_default_content on the other instance. It's a bit messy and you don't know where to look for what. Also create_event changes the state of the Event model, not EventApplication model, and I would really really love to see it moved out from here 💃 🍡

organize/admin.py
```diff
@@ -1,14 +1,155 @@
...
 
 
 class InlineCoorganizerAdmin(admin.TabularInline):
     model = Coorganizer
+    suit_classes = 'suit-tab suit-tab-organizers'
     extra = 1
 
 
+def change_status_to_on_hold(modeladmin, request, queryset):
+    queryset.update(status=ON_HOLD, status_changed_at=timezone.now())
```
bmispelon: You could add a custom `change_status_to()` method on the querysets, so the API between the model and the queryset would be similar.

aniav:
This is massive! 👍 :1st_place_medal: :woman_cartwheeling:

PR #369, opened 2017-01-09
Clone event

asendecka:
If `EventApplication` has `previous_event` set we clone the event content.

- I copy sponors & coaches from old event (`copy_event` command had it broken)
- I don't manage organizers yet - it will be done in separate ticket
- I placed the code in `core/deploy_event.py` - in a separate ticket we will move `create_event` and `add_organizer` as well

### DIFF
--- a/core/deploy_event.py
+++ b/core/deploy_event.py
@@ -11,6 +11,7 @@ def copy_event(previous_event, event_date):
     """
     number = Event.objects.filter(city=previous_event.city,
                                   country=previous_event.country).count()
+    previous_event_id = previous_event.pk
     # If event is already Django Girls City #2, remove #2 from it
     if '#' in previous_event.name:
         generic_event_name = previous_event.name.split(' #')[0]
@@ -29,7 +30,9 @@ def copy_event(previous_event, event_date):
     new_event.save()
 
     # Change url, title and name of previous event to {name} #{number-1}
+    previous_event = Event.objects.get(pk=previous_event_id)
     previous_event.name = previous_event_name
+    previous_event.save()
     previous_event.page_title = previous_event_name
     previous_event.page_url = "{}{}".format(previous_event.page_url, number-1)
     previous_event.save()
--- a/core/deploy_event.py
+++ b/core/deploy_event.py
@@ -1,0 +1,68 @@
+from core.models import Event
+
+
+def copy_event(previous_event, event_date):
+    """
+        Copy event from the previous one. This does the following steps:
+        - change name of previous event to {name} #{number}, i.e.
+          Django Girls London #1
+        - create new event based on previous event
+        - copy content and menu from previous event
+    """
+    number = Event.objects.filter(city=previous_event.city).count()
+
+    # Change the name of previous event to {name} #{number-1}
+    previous_event_name = "{} #{}".format(previous_event.name, number-1)
+    previous_event.name = previous_event_name
+    previous_event.save()
+
+    # Copy event with a name {name} #{number} and new date
+    event_name = "{} #{}".format(previous_event.name, number)
+    new_event = previous_event
+    new_event.pk = None
+    new_event.name = event_name
+    new_event.date = event_date
+    new_event.is_page_live = False
+    new_event.save()
+
+    # Change the title and url of previous event page
+    previous_event.page_title = previous_event_name
+    previous_event.page_url = "{}{}".format(previous_event.page_url, number-1)
+    previous_event.save()
+
+    # populate content & menu from the default event
+    copy_content_from_previous_event(previous_event, new_event)
+    copy_menu_from_previous_event(previous_event, new_event)
+
+    return new_event
+
+
+def copy_content_from_previous_event(previous_event, new_event):
+    """
+        Copies page content from the previous event
+    """
+    previous_event.refresh_from_db()
+    for obj in previous_event.content.all():
+        # fetch related objects before we change pk of the event
+        coaches = obj.coaches.all()
+        sponsors = obj.sponsors.all()
+
+        new_content = obj
+        new_content.id = None
+        new_content.event = new_event
+        new_content.save()
+
+        new_content.coaches = coaches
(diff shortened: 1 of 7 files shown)


core/deploy_event.py
```diff
@@ -0,0 +1,69 @@
...
+
+def copy_event(previous_event, event_date):
+    """
+        Copy event from the previous one. This does the following steps:
+        - change name of previous event to {name} #{number}, i.e.
+          Django Girls London #1
+        - create new event based on previous event
+        - copy content and menu from previous event
+    """
+    number = Event.objects.filter(city=previous_event.city).count()
```
olasitarska: should filter by country as well, there could be two cities of the same name in different countries

core/deploy_event.py
```diff
@@ -0,0 +1,69 @@
...
+
+    previous_event_name = "{} #{}".format(previous_event.name, number)
```
olasitarska: What if `previous_event.name` is already `Django Girls Kraków #3`?

core/deploy_event.py
```diff
@@ -0,0 +1,69 @@
...
+    new_event.pk = None
+    new_event.name = event_name
+    new_event.date = event_date
+    new_event.is_page_live = False
+    new_event.save()
+
+    # Change the title and url of previous event page
+    previous_event.page_title = previous_event_name
+    previous_event.page_url = "{}{}".format(previous_event.page_url, number-1)
+    previous_event.save()
```
olasitarska: Should we handle this save together with `previous_event.save()` in line 19?

core/deploy_event.py
asendecka: Good point :D

organize/models.py
```diff
@@ -104,9 +105,10 @@ def deploy(self):
 
         self.change_status_to(DEPLOYED)
 
-        # TODO: we should recognize here if we should create a new event,
-        # copy old one or copy old and change organizaers.
-        event = self.create_event()
+        if self.previous_event:
```
olasitarska: I think we may need to do more than that here... In the form, we ask user whether they've organized an event before, and if so, we put it in `self.previous_event`. That doesn't mean that there wasn't an event in this city before. I think we should use something like `Event.objects.filter(city=city, country=country).exists()` here.

core/deploy_event.py
```diff
@@ -0,0 +1,69 @@
...
+        coaches = obj.coaches.all()
+        sponsors = obj.sponsors.all()
+
+        new_content = obj
+        new_content.id = None
+        new_content.event = new_event
+        new_content.save()
+
+        new_content.coaches = coaches
+        new_content.sponsors = sponsors
```
olasitarska: I wonder if we should actually copy coaches and sponsors -- it feels like they maybe are very event-specific and it's not often that they all repeat. Better to add them again than copy and assume they are all gonna come back?

organize/models.py
asendecka: Yeah, there is some work to be done here for sure. However, I'm not sure about the suggested solution though. Sometimes you can type city differently in the first edition: for example: Gorzów, Gorzów Wlk., Gorzów Wielkopolski. Once it could be entered without diacritical symbols, latter with these. Also it makes no sense for cities that actually have same name in the same country, i.e.: there are two Józefów in Poland. I know it's a kind of edge case, but we should do more thinking here and we might need info about lat/lng.

Let's create a separate ticket and address it once we actually have latlng properly implemented.

core/deploy_event.py
asendecka: Yeah, it's a fair point. For now I was restoring the logic originally implemented in `copy_event` and it was attempting to do that (but because of the bug it was not actually happening). Happy to remove it here :).

organize/models.py
olasitarska: Yeah, but at least we should create a simple solution here, and then address edgecases later. Now it's just plain wrong?

asendecka: Yeah, I will do a quick update and create a ticket for edge cases 👍

olasitarska:
@aniav @asendecka this is back to you :)

organize/models.py
```diff
@@ -107,9 +108,11 @@ def deploy(self):
 
         self.change_status_to(DEPLOYED)
 
-        # TODO: we should recognize here if we should create a new event,
-        # copy old one or copy old and change organizaers.
-        event = self.create_event()
+        if Event.objects.filter(city=self.city,
+                                country=self.get_country_display()).exists():
+            event = copy_event(self.previous_event, self.date)
```
aniav: We are doing two different things here, we are checking for an event in our database and then using the previous_event from the application object. We should probably check if previous event exists OR if we have a previous event in the database, because in both cases we want to copy contents.

aniav:
@patjouk this is for you now :)

olasitarska:
@aniav added mocking for google accounts tests in #384, think this may fix the issue in tests here?

organize/models.py
```diff
@@ -119,9 +119,19 @@ def deploy(self):
...
+        if self.previous_event:
+            previous_event = self.previous_event
+        else:
+            previous_event = (
+                Event.objects
+                .filter(city=self.city, country=self.country)
+                .order_by("-date")
+                .first()
+            )
+
```
olasitarska: @aniav I think this logic here is wrong, as I discussed with Ola on this pull request earlier. I think we shouldn't be treating "previous_event" as event to copy, as the question asked in the form is "have you organized an event before? if so, choose which". If someone organized an event in Berlin, but then they moved to London, they'd answer this question with "Berlin", but that doesn't mean we should copy the Berlin event with a new date, but we should copy last London event.

olasitarska: Previous discussion for reference: https://github.com/DjangoGirls/djangogirls/pull/369#pullrequestreview-15906366

aniav: Makes sense 👍

olasitarska:
Merging this now as this pull request has already been reviewed. @patjouk if something doesn't work as you want it to work, please create issues :)

Issue #361, opened 2017-01-09
[Core cleanup] Move story to a separate app.


aniav:
Core app has grown a lot in the past months. We shouls move things not related to Event to other apps. While moving it to another app we will have to create migrations that move models to other app while maintaining the data. Django migrations can do that. Ping me for information on how to do that.

This issue is for the Story model and views.

yakupadakli:
https://github.com/DjangoGirls/djangogirls/pull/370

PR #370, opened 2017-01-12
story moved to a separate app

yakupadakli:


aniav:
Hey @yakupadakli we wouldn't like to loose the data from the database when moving the model to the other app, the migration needs to handle that. If you would need help with handling that in migrations, let me know but the basics can be found here: http://stackoverflow.com/questions/25648393/how-to-move-a-model-between-two-django-apps-django-1-7 :)

yakupadakli:
Hi, I updated migrations and I test it twice. It works fine. I used this one: http://stackoverflow.com/questions/25648393/how-to-move-a-model-between-two-django-apps-django-1-7/26472482#26472482

aniav:
Thank you @yakupadakli for your work. This is now merged! :)

#361 Fixed.

PR #371, opened 2017-01-13
Fix reset password url bug

patjouk:
Link was not working ;)

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -86,7 +86,7 @@ class UserChangeForm(forms.ModelForm):
         label="Password",
         help_text="Raw passwords are not stored, so there is no way to see "
                   "this user's password, but you can change the password "
-                  "using <a href=\"password/\">this form</a>.")
+                  "using <a href=\"../password/\">this form</a>.")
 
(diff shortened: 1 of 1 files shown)


core/forms.py
```diff
@@ -86,7 +86,7 @@ class UserChangeForm(forms.ModelForm):
         label="Password",
         help_text="Raw passwords are not stored, so there is no way to see "
                   "this user's password, but you can change the password "
-                  "using <a href=\"password/\">this form</a>.")
+                  "using <a href=\"../password/\">this form</a>.")
```
aniav: Could we use a reverse url here? We should avoid using hardcoded urls, because they might break without our notice :)

patjouk: I tried again and couldn't find how to do it. We need the user ID to generate the URL but I don't know how to get it here.
Do you want to try?

aniav: Checking :)

aniav: @patjouk it seems we were using of Django's Form copied to our source code. I have removed our copy and now the link should be fine :)

PR #375, opened 2017-01-15
Google Apps integration when deploying events

olasitarska:
This integrates @helenst code for Google Apps with our new event deployment.  

Account is created if account with given name doesn't exist. If it exists, we check whether there are common organizers among the old team and the new team. If there are, we do nothing and let them sort out access. Otherwise we migrate old account to a different name and create new one.

Private keys for Google Apps are already setup on PythonAnywhere so this is ready to be deployed, no other configuration needed.


organize/models.py
```diff
@@ -92,6 +93,20 @@ def create_event(self):
 
         return event
 
+    def has_past_team_members(self):
+        """ For repeated events, check whether there are any common
+        team members who applied to organize again
+        """
+        previous_events = Event.objects.filter(city=self.city,
+                                               country=self.get_country_display())
```
aniav: I believe the value stored in the Events database is `self.country`, not `self.get_country_display()`

organize/models.py
```diff
@@ -92,6 +93,20 @@ def create_event(self):
...
+        organizers = []
+        for event in previous_events:
```
aniav: Do we want to filter all previous events or just the last event? Imagine I have handled the first event in city X, for the second event another group of organizers got the access and the e-mail has changed, so now I don't know the password but I still was in the previous organizers.

organize/models.py
```diff
@@ -92,6 +93,20 @@ def create_event(self):
...
+            for organizer in event.team.all():
```
aniav: `organizers = event.team.all().values_list('email', flat=True)`

organize/models.py
```diff
@@ -109,9 +124,20 @@ def deploy(self):
         # copy old one or copy old and change organizaers.
         event = self.create_event()
 
-        # TODO: use method created in separate branch to create gmail acconut
-        # and get password from it.
-        password = "FAKE_PASS"
+        # sort out Gmail accounts
```
aniav: [In a perfect world 😎] I would move the logic of the following few lines to `get_or_create_gmail` or something like this, and let it return the `(email, email_password|None)` tuple.

organize/models.py
```diff
@@ -92,6 +93,20 @@ def create_event(self):
...
+        team members who applied to organize again
+        """
+        previous_events = Event.objects.filter(city=self.city,
+                                               country=self.get_country_display())
+        organizers = []
+        for event in previous_events:
+            for organizer in event.team.all():
+                organizers.append(organizer.email)
+
+        applicants = self.get_all_recipients()
```
aniav: If we are using it here, let's maybe rename to something like `get_organizers_emails`?

core/gmail_accounts.py
```diff
@@ -0,0 +1,95 @@
...
+    """Get the email address for the given slug"""
+    return '%s@djangogirls.org' % slug
+
+
+def create_gmail_account(event):
+    """
+    Create a new account
+    """
+    email = event.email
+    password = get_random_string(length=10)
```
aniav: Will they be required to change the password after first login? If not maybe we could use the password generator from Django :)

helenst: That's a good idea - turns out there's a flag for that: 

```
"changePasswordAtNextLogin": false
```

If that's set in the insert() call then it should prompt them.

See https://developers.google.com/admin-sdk/directory/v1/guides/manage-users#create_user

`get_random_string` is the function that Django uses to generate passwords (`User.make_random_password` is a convenience wrapper around that)

organize/models.py
```diff
@@ -92,6 +93,20 @@ def create_event(self):
...
+        return len(list(set(organizers).intersection(applicants))) > 0
```
helenst: You can do `len(set( .. ))` (no need to convert to list first)

olasitarska:
@aniav @helenst implemented all requested changes :) thanks for the review! 💖

PR #377, opened 2017-01-16
[WIP] Organizers' application form

patjouk:
I've made a few changes:

- I've added a red `*` to the required fields: make it more obvious which fields are mandatory to prevent the risk of people contacting us saying the form is broken because they can't click on the "next" button.
- I've fixed some typos and rewrote some sentences. I've also removed "crazy" because it's an ableist slur.
- I've removed the website slug question: it was too confusing and the error message was horrible. We only change the slug for events happening during a conference: if we received a proposal from a conference, we can ask them if they want a custom slug or not. The slug is auto-generated with the city name.
- I've made "Last name" optional (I need to finish it).

I need to polish it a bit more but I will be happy to have feedback on it ;)


aniav:
I see this is a WIP but I think I might just comment before we go further with that. I believe if we already have the `required=True` on some fields in the backend, we shouldn't copy that information to frontend and hardcode it there. We should use the information from the input to mark it as required. If we keep it in two places we will end up with something marked required in the frontend and not requiring it in the backend. I will be happy to help with that when you finish working on this PR :)

Why do we want to remove the requirement for the last name? I see we have 33 people without last name, out of 660 in the admin :)

organize/models.py
```diff
@@ -30,7 +31,7 @@ class EventApplication(models.Model):
     city = models.CharField(max_length=200)
     country = models.CharField(max_length=200, choices=countries)
     latlng = models.CharField(max_length=30, null=True, blank=True)
-    website_slug = models.SlugField()
+    website_slug = AutoSlugField(populate_from='city', unique=True, editable=True)
```
olasitarska: This shouldn't be unique -- there are gonna be multiple applications that can have the same slug.

templates/organize/form/step1_previous_event.html
```diff
@@ -35,8 +35,8 @@
                     {% endif %}
 
                     <fieldset class="form-section">
-                        <legend class="form-section-heading">Have you organized a Django Girls workshop before?</legend>
-                        <p class="form-section-subhead">If so, we will get you through this process faster as we already know a lot about you! :)</p>
+                        <legend class="form-section-heading, required">Have you organized a Django Girls workshop before?</legend>
```
olasitarska: List of classes shouldn't have commas.

olasitarska: This repeats throughout the whole file, so I won't comment on each.

templates/organize/form/step2_application.html
```diff
@@ -35,7 +35,7 @@
                     {% endif %}
 
                     <section class="form-section">
-                        <label for="{{ wizard.form.about_you.id_for_label }}" class="form-section-heading">Who are you?</label>
+                        <legend for="{{ wizard.form.about_you.id_for_label }}" class="form-section-heading, required">Who are you?</legend>
```
olasitarska: Why are we changing this from label to legend?

olasitarska:
Agreed with @aniav on the use of required. I also added few other minor comments :)

organize/models.py
patjouk: Okay. I'll remove it :)

templates/organize/form/step1_previous_event.html
patjouk: I don't know what happened there :sweat: I'll fix it

templates/organize/form/step2_application.html
patjouk: To keep it coherent with the step1 file.

templates/organize/form/step4_workshop.html
```diff
@@ -68,11 +68,11 @@
 
                     <fieldset class="form-section">
                         <legend class="form-section-heading">When?</legend>
-                        <p class="form-section-subhead">The date of your event should be at least in 1.5 months in the future, but we recommend 3 months for easier planning</p>
+                        <p class="form-section-subhead">There should be <b>at least 1.5 months</b> between now and your event's date, but recommend 3 months for easier planning. If you don't have a precise date for now, you can enter a month and a year.</p>
```
aniav: "but *we* recommend". This part was properly written.

templates/organize/index.html
```diff
@@ -14,7 +14,7 @@
             <section class="col-xs-12 col-md-8">
                 <p>Django Girls events are organized by groups of wonderful volunteers in cities all over the world. Events are always non-profit and free for participants. We do not pay coaches, speakers or organizers. Attendees don’t need any previous knowledge about programming and there are no age limitations. All attendees need is a laptop and some curiosity!</p>
 
-                <p>Until today, <span>{{ organizers_count }}</span> volunteers organized <span>{{ all_events_count }}</span> Django Girls workshops. Join them and bring Django Girls to your city!</p>
+                <p>Since July 2014, <span>{{ organizers_count }}</span> volunteers organized <span>{{ past_events_count }}</span> Django Girls workshops. Join them and <a href="/organize/prerequisites/">bring Django Girls to your city!</a></p>
```
aniav: The url should be a `{% url 'organize:prerequisites' %} AFAIR.

templates/organize/form/step4_workshop.html
patjouk: fixed

templates/organize/prerequisites.html
```diff
@@ -55,7 +55,7 @@
                         </fieldset>
 
                         <fieldset>
-                            <legend>Confirm that you meet following requirements to organize a workshop:</legend>
+                            <legend class="required">Confirm that you meet following requirements to organize a workshop:</legend>
```
aniav: The "required" here should be removed too.

organize/models.py
```diff
@@ -182,7 +183,7 @@ class Coorganizer(models.Model):
         related_name="coorganizers")
     email = models.EmailField()
     first_name = models.CharField(max_length=30)
-    last_name = models.CharField(max_length=30)
+    last_name = models.CharField(max_length=30, blank=True)
```
aniav: We have to remember to check if everything is right after adding blank here. We have methods taking last_name as something that should just exist because it was required.

aniav: Also if we permit `blank=True`, and we don't want `null=True` (as this is in #83 ) we should add `default=""` here.

templates/organize/index.html
patjouk: fixed

templates/organize/prerequisites.html
patjouk: fixed

aniav:
I have checked what's with the `<legend>` and in the first step it was used because it is a definition of a label for many fields. That's how it should be used, looking into [w3schools](http://www.w3schools.com/tags/tag_legend.asp), so I believe we would like label to be back on the files where they were in most cases.

templates/organize/index.html
```diff
@@ -32,7 +32,7 @@
             </section>
 
             <section class="col-xs-12 col-md-4 sidebar">
-                <a href="/organize/prerequisites/" class="cta">Apply to organize a workshop&nbsp;»</a>
+                <a href="{% url "organize:prerequisites" %}" class="cta">Apply to organize a workshop&nbsp;»</a>
```
aniav: 👍

aniav:
Closing this PR in favour of 378

PR #385, opened 2017-01-22
Add saving coordinates of the city when creating EventApplication

olasitarska:


olasitarska:
This build needs to be refreshed when #384 is merged

PR #391, opened 2017-01-24
[WIP] Add event number field

aniav:
We are adding a new field on the `Event` model which will indicate the number of the event in the city it happens in. This will help create proper links and titles on the old event pages, as well as adding links to previous events.


patjouk:
Hi @aniav :)
Where are you on this? This bugs is still super annoying and it would help me a lot if it got fixed ;)

patjouk:
I've fixed the bug for the 'copy_event' management command (#423).

aniav:
I don't think I will have time to get back to this.

PR #393, opened 2017-01-25
Set a random cover picture for new events

patjouk:
When created, new events automatically get a random cover picture. We can now add the event directly to the event list :tada:


patjouk:
Don't merge it for now: I'll change how we handle pictures to prevent reupload them all the time.

patjouk:
Closing this PR for a better one :wink:

PR #394, opened 2017-01-26
Removed permissions from sample data

madjar:
The permissions in the files are out-of-date, and the permissions are generated by the `migrate` command anyway.


madjar:
Apparently, I can't reproduce the problem anymore.

PR #404, opened 2017-02-02
Add core.gmail_accounts to codecov ignores

madjar:
Since this file is not tested for pull requests, codecov falsely reports a
decreased coverage for every single pull request. Ignoring this file restores
meaningful reports in pull requests.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -14,3 +14,4 @@ coverage:
     - tests/*
     - migrations/*
     - djangogirls/wsgi.py
+    - core/gmail_accounts.py


aniav:
I am not sure what to do with this pull request. We are not testing this file, but I think we definitely should have tests for it as it's part of an important functionallity. Why do you think this should be added to ignores?

madjar:
Sorry, I forgot to give the context for this. There it is:

My previous pull request got a failure from codecov, because it reduces the coverage (except it doesn't). The report from codecov, which can be seen there (https://codecov.io/gh/DjangoGirls/djangogirls/compare/3e4b5174428186d0255981281365ecd8c5bf634a...d33998c458c285cc5858281fcef9851c79bb0041), show that the loss of coverage is in `core/gmail_accounts`.

This is because [the tests are only run](https://github.com/DjangoGirls/djangogirls/blob/master/core/tests/test_gmail_accounts.py#L23) when the `GAPPS_PRIVATE_KEY` is set, and that is only the case when travis is testing master, not pull requests (this is a security mesure, to prevent someone from submitting a PR that prints the private key to steal it).

So the code is actually tested, but only for the master branch. This has the annoying side effect that codecov considers that every single pull request removes the coverage from the file. By adding it to the ignores, we remove this false positive report for all the pull requests.

PR #411, opened 2017-02-20
events are editable until their date has passed

### LINKED ISSUE #398: Admin considers current events "past" and won't allow editing
This bit me when I wanted to update our page the morning of the event. The admin still showed all the pages, but wouldn't allow editing them; the fields were displayed as static text instead of inputs. Luckily I got around this because the admin still allowed creating completely new pages and changing their order, I just had to get it all right in one shot.

The admin should allow editing until at least the day *after* the day of the event.

davidism:
closes #398

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -178,13 +178,9 @@ class Event(models.Model):
         verbose_name_plural = "List of events"
 
     def is_upcoming(self):
-        if not self.date:
-            return False
         now = timezone.now()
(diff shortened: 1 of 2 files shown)


davidism:
Coverage is technically lower because more lines were removed than added. All relevant lines are still covered.

aniav:
Thanks! :)

PR #426, opened 2017-03-11
Restrict Event and Form admin list_filters by user

Ian-Foote:
If a user is not a superuser, only allow filtering by forms or events for any team they are a member of.

### DIFF
--- a/applications/admin.py
+++ b/applications/admin.py
@@ -54,10 +54,17 @@ class FormAdmin(admin.ModelAdmin):
     get_submissions_url.short_description = "Applications"
 
 
+class FormFilter(admin.SimpleListFilter):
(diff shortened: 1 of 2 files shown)


Ian-Foote:
I didn't manage to get a properly working local dev environment, so I haven't manually tested these actually work. If someone can confirm they do or don't, I'd appreciate it.

I'm also not sure if this needs tests, and what exactly they should look like if it does.

aniav:
@Ian-Foote Thanks! 👍  I managed to merge it. Can you tell me what problems did you have with setting the local dev env? It's an important feedback for us because we probably should fix something on our end if bringing the project up isn't easy.

PR #431, opened 2017-05-07
Add missing migration

ulgens:
After a fresh install and applying all existing migrations, still there is a difference between database and models. It's the result of missing migration, which comes with this PR :)


olasitarska:
Thank you! :)

PR #441, opened 2017-09-06
Add Russian version of CoC

nazarov-tech:
We ([djangogirls.org/spb](djangogirls.org/spb)) carefully translated Code of Conduct into Russian.

I tested it locally. Please, accept :)


nazarov-tech:
there are Django Girls in two biggest cities of Russia now: both Moscow and St. Petersburg.

So yeah, we do need a Russia version of CoC.

PR #446, opened 2017-11-04
Rename description block in registration templates to prevent name collision - Fix #445

### LINKED ISSUE #445: description block with HTML tags break the site's rendering
On some pages, the `description` block contains HTML tags. This is problematic because it is used inside of `<meta name="description" content="">`, which expects its value to be a string. Here is an example of the resulting invalid HTML:

```html
<meta name="description" content="<p>We just sent reset instructions to your email. If you have any issues, <a href=/account/password_reset/>try again</a> and if you still don't get anything, <a href="/contact/">say hello</a>!</p>">
```

I spotted this issue in `password_reset_done` and `password_reset_complete`: https://github.com/DjangoGirls/djangogirls/blob/0730e5a5ddffea1ab830074aaf2968c5a4f091d7/templates/registration/password_reset_done.html#L5-L7

Screenshots:

djangogirls_password_reset_done
djangogirls_password_reset_complete

------------

Trying to identify why this is happening, this seems to be specific to the `registration` templates. They are (incorrectly?) reusing the `description` block as a way to generate the main content of the page:

https://github.com/DjangoGirls/djangogirls/blob/a4c9c5489dc74dbe2d9ceeaf9ee06857bc40b7b4/templates/registration/base.html#L7-L9

If that's indeed the problem, then the solution is to introduce a new block on all of those templates instead of recycling `description`. It would also be worth looking around to make sure those are the only templates that have this issue.

I'm happy to submit a PR if that's helpful.

thibaudcolas:
Fixes #445.

This is just a search and replace following @lightstrike's suggestion. I tried it on all the password reset steps, all seems well.

### DIFF
--- a/templates/registration/base.html
+++ b/templates/registration/base.html
@@ -5,7 +5,7 @@
     <div class="row centered">
         <div class="col-md-6 col-md-offset-4">
             <h2>{% block title %}{% endblock title %}</h2>
-            <p>{% block description %}{% endblock description %}</p>
+            <p>{% block registration_message %}{% endblock registration_message %}</p>
             <form action="" method="post">
                 {% csrf_token %}
                 {% for field in form %}
--- a/templates/registration/password_reset_complete.html
+++ b/templates/registration/password_reset_complete.html
@@ -2,6 +2,6 @@
 
 {% block title %}Password Reset Complete{% endblock title %}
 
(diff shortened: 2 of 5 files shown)


aniav:
This was a really silly bug! Thank you! Merging 👍

PR #452, opened 2017-12-23
Change website contents min-width so CodeMirror does not overflow

### LINKED ISSUE #434: Admin: Website contents not responsive on tablet
I'm trying to edit the website contents (values) on the website, but I can't save the modifications because the text box overlaps the save button. Since I'm using a tablet, I can't just drag the corner to change the size of the text box. I believe a simple change on the template would fix this.

I'm attaching a screenshot of how it's displayed. Let me know if you need any more info.

img_1933

thibaudcolas:
Fixes #434. I went for the simpler solution – make the page scroll horizontally – because it's the simplest behaviour, and there is already a 960px min-width on the body for other pages so it's not a big difference.

### DIFF
--- a/static/source/css/admin.styl
+++ b/static/source/css/admin.styl
@@ -1,3 +1,7 @@
+// Make sure there is enough space for the fixed-width CodeMirror editor.
+.core-eventpagecontent
+    min-width: 1351px
+
 .save-box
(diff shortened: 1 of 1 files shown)


aniav:
Thanks! 🙃

PR #464, opened 2018-03-29
Added crowdfunding donors page

amakarudze:
Added a page to list all our crowdfunding donors. The link to the page is on Support Us dropdown. Donors can be added to the database using the ```./manage loaddata donors.json``` command. The json file does not contain the "visible" field so that will have to be entered manually based on the donors preferences in the admin section for a donor to appear on the page.

Ping me for any questions.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -5,11 +5,13 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.template import TemplateDoesNotExist
 from django.utils import timezone
 from django_date_extensions.fields import ApproximateDate
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 
 from patreonmanager.models import FundraisingStatus
 
 from .models import Event, User
 from story.models import Story
+from sponsor.models import Donor
 
 
 def index(request):
@@ -143,3 +145,20 @@ def coc(request, lang=None):
         return render(request, template_name)
     except TemplateDoesNotExist:
         raise Http404("No translation for language {}".format(lang))
+
+
+def crowdfunding_donors(request):
+    donor_list = Donor.objects.filter(visible=True).order_by('amount')
+    paginator = Paginator(donor_list, 60)  # Show 60 donors per page
+    page = request.GET.get('page')
+    try:
+        donors = paginator.page(page)
+    except PageNotAnInteger:
+        # If page is not an integer, deliver first page.
+        donors = paginator.page(1)
+    except EmptyPage:
+        # If page is out of range (e.g. 9999), deliver last page of results.
+        donors = paginator.page(paginator.num_pages)
+    return render(request, 'core/crowdfunding_donors.html', {
+        'donors': donors,
+    })
--- a/donors.json
+++ b/donors.json
@@ -1,0 +1,1186 @@
+[
+  {
+  "pk": 1,
+  "fields": {
+    "name": "Danielle Topping",
+    "amount": 75
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 2,
+  "fields": {
+    "name": "Annette McCullough",
+    "amount": 20
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 3,
+  "fields": {
+    "name": "David Barragán Merino",
+    "amount": 20
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 4,
+  "fields": {
+    "name": "Alja Isaković",
+    "amount": 75
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 5,
+  "fields": {
+    "name": "Meri Williams",
+    "amount": 300
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 6,
+  "fields": {
+    "name": "Rebecca M Long",
+    "amount": 50
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 7,
+  "fields": {
+    "name": "Oliver Andrich",
+    "amount": 50
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 8,
+  "fields": {
+    "name": "Patrick Arminio",
+    "amount": 20
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 9,
+  "fields": {
+    "name": "Aleksandra Kunysz",
+    "amount": 15
+  },
+  "model": "sponsor.donor"
+},
+  {
+  "pk": 10,
+  "fields": {
+    "name": "HackSoft",
+    "amount": 300
+  },
+  "model": "sponsor.donor"
+},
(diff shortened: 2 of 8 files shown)


core/views.py
```diff
@@ -143,3 +145,20 @@ def coc(request, lang=None):
         return render(request, template_name)
     except TemplateDoesNotExist:
         raise Http404("No translation for language {}".format(lang))
+
+
+def crowdfunding_donors(request):
+    donor_list = Donor.objects.filter(visible=True).order_by('amount')
```
asendecka: Should we add secondary order by? I.e. name? Django by default orders ascending, so order `amount` will return smaller values first. I think it makes sense to display bigger donors first, right?

Like: `order_by('-amount', 'name')`

sponsor/models.py
```diff
@@ -24,3 +24,16 @@ def logo_display_for_admin(self):
             return "No logo"
 
     logo_display_for_admin.allow_tags = True
+
+
+class Donor(models.Model):
```
asendecka: We added a new model, but the PR does not contain the migration for the new model. We do that with `./manage.py makemigrations` :).

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,49 @@
...
+                We managed to raise a total of $13,314!</p>
+
+            <p>If you missed our campaign but will still like to support our work, you can donate to us via
+                PayPal or credit/debit card for once off donations. Alternatively, you can sign up on <a href="http://patreon.com/djangogirls">Patreon
+                </a> for monthly donations. Thank you for supporting our work.</p>
+
+            <p>Thank you. Your support means the world to us.</p>
+        </div>
+        <div class="col-md-6">
+                    <img class="img-responsive center-block" alt="Django Girls Gliwice #1." src="{% static 'img/global/donate/tnr.jpg' %}">
```
asendecka: I seem not to have this picture displayed properly.

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,49 @@
...
+                </div>
+        </div>
+
+        <div class="row">
+            <h2>They Support Us</h2>
+            {% for donor in donors %}
+                 {% if forloop.first %}<div class="row">{% endif %}
+                <div class="col-md-3">
+                    <p>{{ donor.name }}</p>
```
asendecka: I wonder if we should display amount or not? What do you think?

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,49 @@
...
+            <p>Thank you. Your support means the world to us.</p>
+        </div>
+        <div class="col-md-6">
+                    <img class="img-responsive center-block" alt="Django Girls Gliwice #1." src="{% static 'img/global/donate/tnr.jpg' %}">
+                </div>
+        </div>
+
+        <div class="row">
+            <h2>They Support Us</h2>
+            {% for donor in donors %}
```
asendecka: I think we need to add some styles to make donors look a bit nicer: now it's just a simple columns with donor's names. Maybe we could add our Django Girls heart near each name? Or choose some nice font?

templates/global/menu.html
```diff
@@ -27,9 +27,10 @@
                  <li class="dropdown">
           <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Support us 💕 <span class="caret"></span></a>
           <ul class="dropdown-menu">
-            <li><a href="{% url 'core:donate' %}">Donate</a></li>
-            <li><a href="https://store.djangogirls.org/">Buy a t-shirt</a></li>
-            <li><a href="https://www.patreon.com/djangogirls">Our Patreon page</a></li>
+              <li><a href="{% url 'core:donate' %}">Donate</a></li>
+              <li><a href="https://store.djangogirls.org/">Buy a t-shirt</a></li>
+              <li><a href="https://www.patreon.com/djangogirls">Our Patreon page</a></li>
+              <li><a href="{% url 'core:crowdfunding-donors' %}">Our Supporters</a></li>
```
asendecka: I wonder if this not too hidden on the page?

templates/core/index.html
```diff
@@ -42,37 +42,6 @@ <h2>What is Django Girls?</h2>
                 </div>
             </div>
```
asendecka: Referring to my comment about donors page being a bit hidden in the menu: maybe we could leave it there, but have a section on the main page that features our donor's page a bit, so people could discover it easier - I think it would be nice for our donors to have that featured for some time. So we would need a section with a sentence or two about campaign and how greateful we are to our donors, a nice picture and link to the donors page. 

If we feel fancy, we could even fetch a couple of donor's names and randomly feature them on the main page?

core/views.py
```diff
@@ -143,3 +145,20 @@ def coc(request, lang=None):
...
+    paginator = Paginator(donor_list, 60)  # Show 60 donors per page
```
asendecka: Let's remove completely paginator. We want all our donors on the same page.

donors.json
```diff
@@ -0,0 +1,1186 @@
+[
```
asendecka: Are all these names of the current sponsors? Are they all visible and agreed to have their names listed with the amount? If not, we should remove this file completely from the repo and add it through Django admin.

asendecka:
To meet criteria for test coverage, we need tests for following things:
- we need to add pytest fixtures here: https://github.com/DjangoGirls/djangogirls/blob/master/tests/conftest.py - one called `visible_donors` with a couple of donors with visibility set to `True` and another one called `hidden_donors` with visibility set to `False`

Something like:
```
@pytest.fixture()
def visible_donors():
    donors = Donor.objects.bulk_create([
        Donor(name="Ola", amount="50", visible=True),
        [ADD ONE OR TWO MORE DONORS HERE]
    ])
    return donors

@pytest.fixture()
def hidden_donors():
    hidden_donors = Donor.objects.bulk_create([
        Donor(name="Mateusz", amount="50", visible=False),
       [ADD ONE OR TWO MORE DONORS HERE]
    ])
    return hidden_donors
```
- in `/core/test_views.py` we want to add: `test_crowdfunding_donors` which looks similar to the test for `index` (here: https://github.com/DjangoGirls/djangogirls/blob/master/tests/core/test_views.py#L8) that tests `core:crowdfunding-donor` url. We would be able to pass pytest fixtures we created in a previous step as a arguments to the test function like this:

```
def test_crowdfunding_donors(client, visible_donors, hidden_donors):
    # Access crowdfunding donor page
    resp = client.get(reverse('core:crowdfunding-donors'))
    assert resp.status_code == 200

    # Check if it returns a list of past and future events
    assert 'donors' in resp.context

    # Only the visible donors
    donor_ids = set(donor.pk for donor in resp.context['donors'])
    visible_donors = set(visible_donor.pk for visible_donor in visible_donors)
    assert donor_ids == visible_donors
```

PR #466, opened 2018-05-08
New changes to donors page

amakarudze:


core/views.py
```diff
@@ -5,11 +5,13 @@
 from django.template import TemplateDoesNotExist
 from django.utils import timezone
 from django_date_extensions.fields import ApproximateDate
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
```
asendecka: We don't need paginator anymore, let's remove this import.

djangogirls/settings.py
```diff
@@ -276,3 +276,15 @@
 GAPPS_ADMIN_SDK_SCOPES = 'https://www.googleapis.com/auth/admin.directory.user'
 GAPPS_PRIVATE_KEY_ID = os.environ.get('GAPPS_PRIVATE_KEY_ID', '')
 GAPPS_PRIVATE_KEY = os.environ.get('GAPPS_PRIVATE_KEY', '')
+
+DATABASES = {
```
asendecka: We need to remove this bit - it contains your local password and we have a proper db configuration earlier in the file.

sponsor/migrations/0006_auto_20180503_2310.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
asendecka: There are so many migrations here, would be better to have only one, instead of changing it multiple times. Could you remove all around `Quote` and generate one that is clean?

sponsor/models.py
```diff
@@ -24,3 +24,29 @@ def logo_display_for_admin(self):
...
+
+    def __str__(self):
+        return self.name
+
+
+    class Meta:
+        ordering = ("amount",)
+
+
+class Quote(models.Model):
```
asendecka: Let's remove `Quote`

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,45 @@
...
+                 {% if forloop.first %}<div class="row">{% endif %}
+                <div class="col-md-4">
+                    <h4 class="donor-text">
+                        <img src="{% static 'img/global/donate/pixel-heart.png' %}"> {{ donor.name }}</h4>
+                </div>
+                {% if forloop.counter|divisibleby:3 %}</div><div class="row">{% endif %}
+                {% if forloop.last %}</div>{% endif %}
+                {% if forloop.counter|divisibleby:6 %}</div><div class="row"><br />
+                    <div class="col-md-9">
+                     <blockquote>{% for quote in quotes %}{{ quote.quote }} <p>{{ quote.name }}</p>
```
asendecka: We should move `{{ quote.name }}` to the `cite` in `blockquote` as specified here: https://www.w3schools.com/tags/tag_blockquote.asp

So it will look like:
```
<blockquote cite='{{ quote.name }}'>{% for quote in quotes %}{{ quote.quote }}</blockquote>
```

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,45 @@
...
+        </div>
+        </div>
+
+        <div class="row">
+            <h2>They Support Us</h2>
+            {% for donor in donor_list %}
+                 {% if forloop.first %}<div class="row">{% endif %}
+                <div class="col-md-4">
+                    <h4 class="donor-text">
+                        <img src="{% static 'img/global/donate/pixel-heart.png' %}"> {{ donor.name }}</h4>
```
asendecka: Let's remove the heart here.

sponsor/models.py
```diff
@@ -24,3 +24,29 @@ def logo_display_for_admin(self):
...
+    name = models.CharField(max_length=200, null=True, blank=True)
+    quote = models.TextField()
+    picture = models.FileField()
```
asendecka: If we really want a picture (but I'm fine with removing it completely and just have a text & name in Quote alone), we could do that similarly to the: https://github.com/amakarudze/djangogirls/blob/d951efb7ae05548793a5b38cb7c3f00116b9218c/core/models.py#L118 using `ImageField`.

But to be fair, let's just remove it completely and not worry about picture - this will be extra work to choose them and make sure they are displayed nicely. I think we could just skip this step and prioritise having the page up and running.

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,45 @@
...
+                </div>
+                {% if forloop.counter|divisibleby:3 %}</div><div class="row">{% endif %}
+                {% if forloop.last %}</div>{% endif %}
+                {% if forloop.counter|divisibleby:6 %}</div><div class="row"><br />
```
asendecka: Let's have more names before each quote - let's say 24 to have a big chunk with names.

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,38 @@
...
+
+        <div class="row">
+            <h2>They Support Us</h2>
+            {% for donor in donor_list %}
+                 {% if forloop.first %}<div class="row">{% endif %}
+                <div class="col-md-4">
+                    <h4 class="donor-text">{{ donor.name }}</h4>
+                </div>
+                {% if forloop.counter|divisibleby:3 %}</div><div class="row">{% endif %}
+                {% if forloop.last %}</div>{% endif %}
```
asendecka: Ok, so what I propose for the quotes is the following:

1. Add file `quotes.py` under `core` directories and add hardcoded quotes like: 

```
DONORS_QUOTES = [
    {'text': '', 'author': '', 'url': ''},
    ...
]
```

and add a couple of quotes (with author and link if we could find a link to the tweet or something).

2. In `core/views.py` we should import this
```
from .quotes import DONOR_QUOTES
```
and later in the `crowdfunding_donors` function we could send it to the template:

```
return render(request, 'core/crowdfunding_donors.html', {        
    'donor_list': donor_list,
    'quotes': DONOR_QUOTES,
})
```

3. Create a `element_by_counter.py` file under `core/templatetags` with something like this:

```
from django import template
register = template.Library()

@register.simple_tag
def element_by_counter(data, counter, divider):
     index = int(counter / divider)
     if index < len(data):
          return data[index]
     return ''
```
4. In template add the file name to the `{% load %}` tag.

5. In template let's use it to display quote based on the counter we get from the outside loop.

```
{% if forloop.counter|divisibleby:XX}
    {% element_by_counter quotes forloop.counter XX as quote %}
    {{ quote.text }} <--- [let's use quote.text, quote.author and quote.url to style it here nicely]
{% endif %}
```

We should add tests of the new tag, too.

tests/conftest.py
```diff
@@ -4,6 +4,7 @@
 
 from core.models import Event, User
 from pictures.models import StockPicture
+from sponsor.models import Donor, Quote
```
asendecka: Tests are breaking because there is a missing migration file. If you run `./manage.py makemigrations` it should add Donor model.

asendecka: As you removed `Quote` completely, you should remove import here, too.

tests/core/test_views.py
```diff
@@ -122,3 +122,18 @@ def test_coc_redirect(client):
...
+    # Access crowdfunding donors page
+    resp = client.get(reverse('core:crowdfunding-donors'))
+    assert resp.status_code == 200
+
+    # Check if it returns list of donors in response
+    donors = visible_donors
+    assert donors in resp.context
+
+    # Only the visible donors
+    donor_ids = set(donor.pk for donor in resp.context['donors'])
```
asendecka: I believe in context we get `donor_list` instead of `donors`. We should use correct name.

aniav:
Do we still want to work on this and merge this? 👍

aniav:
Closing due to inactivity 🙌

PR #467, opened 2018-05-22
remove Opbeat 😢

beniwohli:
As you might have heard, Opbeat was acquired by Elastic a year ago, and on May 24 (this Thursday), Opbeat.com will finally shut down 😢

I removed all references to `opbeat` from the code base. This should ensure that no interruptions of service will occur once Opbeat.com is gone (there shouldn't be any catastrophic side effects even if you don't deploy this before May 24).

If y'all are interested, I can check if we can hook you up with a sponsored instance of Elastic APM, the product that the Opbeat team has been working on since joining Elastic. It comes with many of the same features as Opbeat (but not all, yet, we're working on that :) ).

### DIFF
--- a/core/management/commands/add_organizer.py
+++ b/core/management/commands/add_organizer.py
@@ -3,7 +3,6 @@ import djclick as click
 
 from core.forms import AddOrganizerForm
 from core.models import Event
-from core.utils import opbeat_logging
 
 DELIMITER = "\n-------------------------------------------------------------\n"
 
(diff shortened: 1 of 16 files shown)


aniav:
😢😢😢

We should definitely find something else to check the website.

beniwohli:
@aniav we @elastic would love to sponsor an Elastic APM hosted instance, unfortunately the release of our cloud offering is still a few months out. I can get back to you once we have a clearer idea when it will be available.

aniav:
👍 for now I've just setup sentry.io. To be honest our usage of error reporting is veeery basic, so I think sentry should be more than enough :)

PR #469, opened 2018-06-02
Only link coach name if there is a url

tbrisker:


tbrisker:
@ramonsaraiva thanks for taking a look, I've updated the PR to combine the check with the similar check being done for the image so it's clearer now imho what markup will be used with or without url.

aniav:
Thanks! 💃 💃 💃

PR #471, opened 2018-06-03
Fix typo: make "Services" plural

das-g:
make "Services" plural in "I acknowledge that some of my data will be used on Third Party Sites and Service**s**." question


das-g:
Why doesn't the `master` branch have coverage data available? If coverage is being measured automagically, shouldn't that also happen on the repo's default branch?

PR #474, opened 2018-06-26
Add crowdfunding donors page to website

amakarudze:
We promised our crowdfunding supporters mention of their names in newsletter and on our website. These changes add a page to the website with the list of our awesome crowdfunding supporters who chose to be visible donors during the campaign.


core/templatetags/element_by_counter.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+register = template.Library()
+
+@register.simple_tag
+def element_by_counter(data, counter, divider):
+     index = int(counter / divider)
```
asendecka: Let's make sure we handle `divider = 0` well. If it's `0` we should return empty string, too, instead ending up with division by zero error.

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,50 @@
...
+                <img src="{% static 'img/global/donate/django_crowdfunding_heart.png' %}"></a>
+            </div>
+        </div>
+
+        <br/>
+
+        <div class="row">
+            <h2>Our awesome supporters ✨</h2>
+            {% for donor in donor_list %}
+                 {% if forloop.first %}<div class="row">{% endif %}
```
asendecka: This line is not indented nicely - there is one extra space I think. It's super small thing, but it's nice to have nicely formatted code :D. Let's remove one space before `{%`

templates/includes/_main_sponsors.html
```diff
@@ -104,4 +104,14 @@
             <img class="img-responsive center-block" src="{% static 'img/global/supporters/wsp.png' %}" alt="WSP Digital" />
         </a>
     </div>
+    <div class="col-md-3">
```
asendecka: Is there any reason we add extra sponsors here? Are they sponsors who donated through Crowdfunding?

core/templatetags/element_by_counter.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+register = template.Library()
+
+@register.simple_tag
+def element_by_counter(data, counter, divider):
```
asendecka: Could you add tests to this template tag? Could be in separate PR as we don't want to block this to be merged.

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,50 @@
...
+            <div class="col-md-6">
+                <p>
+                This year (2018) we ran a crowdfunding campaign on Indiegogo's generosity website from
+                the 29th January to the 29th of March with the aim of raising US $10,000. We are truly
+                grateful to all our awesome supporters who helped reach our target goal. We managed to
+                raise a total of $13,314 - 133% of our goal!
+                </p>
+            </div>
+            <div class="col-md-6">
+                <img src="{% static 'img/global/donate/django_crowdfunding_heart.png' %}"></a>
```
asendecka: Is this image added to the site?

templates/core/crowdfunding_donors.html
```diff
@@ -0,0 +1,50 @@
...
+            {% for donor in donor_list %}
+                 {% if forloop.first %}<div class="row">{% endif %}
+                <div class="col-md-4">
+                    <h4 class="donor-text">{{ donor.name }}</h4>
+                </div>
+                {% if forloop.counter|divisibleby:3 %}</div><div class="row">{% endif %}
+                {% if forloop.last %}</div>{% endif %}
+                {% if forloop.counter|divisibleby:42 %}
+                    <br />
+                    {% element_by_counter quotes forloop.counter 42 as quote %}
```
asendecka: `forloop.counter` starts from `1`, not `0` - that means that first quote will never be displayed. Instead we should use `forloop.counter0` (but only in this line!)

templates/includes/_main_sponsors.html
amakarudze: Yes, they are new sponsors from crowdfunding campaign as well as new sponsors who came on board after the campaign.

core/templatetags/element_by_counter.py
amakarudze: Will work on this. Thanks

core/templatetags/element_by_counter.py
amakarudze: Done, thanks.

templates/core/crowdfunding_donors.html
amakarudze: Fixed indentation.

PR #485, opened 2018-08-29
Save new coaches to the database

mUtterberg:
screen shot 2018-08-29 at 8 03 53 am

There is currently an issue in which coaches are not saving to my event. I notice that class Coach does not have a self.save anywhere in the model definition and the views.py for this function is an empty file. Will this change allow my coach entries to save in the database?

### DIFF
--- a/coach/models.py
+++ b/coach/models.py
@@ -43,3 +43,6 @@ class Coach(models.Model):
                 return DEFAULT_COACH_PHOTO
 
         return DEFAULT_COACH_PHOTO
+    
+    # Adding new coaches currently results is "saved" confirmation, but no entry in the database
+    self.save()


aniav:
Hey @mUtterberg it seems the change is breaking the build on Travis CI. Can you check if the change is actually valid?

mUtterberg:
@aniav I'm actually not entirely sure it's a valid change! I'm just a little bit at a loss as to why the coaches I am saving via the event admin site are not showing in the coaches database. The pull request can probably be closed because there's an open issue, but I don't think I can figure out where to fix the bug!

aniav:
I think I remember what the issue is: so the coach is actually being saved in the DB, but it's not being assigned to any event, so effectively you need to do add them separately. Not sure if this is still true, as I haven't looked into the code for quite some time now :)

das-g:
So maybe the first task should be to write a (for now failing) automated test that shows what the issue is. (And can be used to tell when it's fixed and be kept to avoid regressions afterwards.)

aniav:
I don't do much with the code here but you can assign me to whatever you would need a reviewer for 👍

PR #490, opened 2018-09-24
Change Payment queryset to be a manager

ramonsaraiva:
It is good to keep all data manipulation in a manager, so querysets can naturally behave being chainable filter/query level manipulation.


aniav:
Tested locally 👍

Issue #454, opened 2018-01-30
Don't block event deployment when we can't send an e-mail to one of the organizers


aniav:
This is a bug we currently have. Details below:

https://opbeat.com/django-girls/djangogirlsorg/errors/88/

ramonsaraiva:
Hey @aniav could you add more context to this since `opbeat` is gone?

aniav:
Oh, I didn't get Opbeat is going to be gone. When an application event is being deployed we will create a webpage for it, create the accounts for organizers and send the account details to them. Sometimes the organizers' e-mails are spelled wrong causing the sending mechanism to fail. The code should take this situation into account so that failed e-mail sending doesn't cause the whole deployment to fail, leaving the deployment state unclear.

Thanks for going through the issues here 👍

PR #491, opened 2018-09-29
Move deployment email logic out of atomic transaction

ramonsaraiva:
Closes https://github.com/DjangoGirls/djangogirls/issues/454

The deployment e-mails were being sent in the same function that deploys an event, but `transaction.atomic` decorates this function, so in case the e-mails somehow fail to be sent, the whole deployment will fail.

This moves the e-mail logic out of the atomic function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 *.sqlite3
 *.pyc
+.python-version
 staticfiles/
 uploads/
 local_settings.py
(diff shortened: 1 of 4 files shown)


ramonsaraiva:
`def view_change_application_status` is not yet covered by unit tests, that's the reason of codecov failing.

Let me know if it is needed for this, or if i can add the tests in future PRs.

PR #492, opened 2018-10-03
Organizers can see other workshops' pages that aren't live when logged in

### LINKED ISSUE #465: Organizers can see other workshops' pages that aren't live when logged in
How I noticed this:
- I was logged in as Django Girls Santa Maria organizer.
- I typed "djangogirls.org/brasilia" in the browser, which is an event that I know it will happen in four months
- It returned the template page for Brasilia, with the message "Hi there! You can see this page because you're logged in admin. When you want to make it public, go to your website section in the admin and click on 'Is live'."

This probably shouldn't happen. Not "live" pages should be shown only to it's own organizer's and to Django Girls staff/admin.

ramonsaraiva:
Users will only be able to preview an event page if they are superusers or organizers.

Closes #465

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -205,6 +205,13 @@ class Event(models.Model):
                    for x in self.team.all()]
         return ", ".join(members)
 
+    def has_organizer(self, user):
+        """
+        Return whether a specific user is an organizer of the event
+        """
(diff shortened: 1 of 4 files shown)


aniav:
Tested locally, works like it's supposed to 👍 🎉

Issue #435, opened 2017-06-26
settings.STORE_DISCOUNT_CODE is not set


magul:
See:
```
Discount in our store
Use the code "None" to get a 15% discount on your order in our store. You can share this code with your coaches and attendees. All proceeds will go to Django Girls Foundation (non-profit organization). 
```

aniav:
This is now fixed (better late then never 👍 )

PR #500, opened 2018-12-11
Adding django-dotenv and example .env file to handle missing configuration

apoclyps:
### What's Changed

As a solution to #435, the following pull request attempts to reduce the complexity of managing secrets locally by providing an example `.env` file containing all secrets that are loaded as environmental configuration with some sensible defaults during development.

### Technical Description

During development, a `.env` file is loaded on the developer's machine by copying the example file and providing secrets within the `.env`. 

~Adding the `LOCAL_DEVELOPMENT` flag will load the `.env` configuration; by default, this will not be loaded. This is achieved by adding the following two lines to the setup process.~ This has since been removed due to feedback, instead the `.env` file will always be loaded (if provided).

```sh
cp .env.example .env
```

> Adds django-dotenv as a dependency


djangogirls/settings.py
```diff
@@ -1,8 +1,14 @@
 import os
 import dj_database_url
+import dotenv
 import sentry_sdk
 from sentry_sdk.integrations.django import DjangoIntegration
 
+LOCAL_DEVELOPMENT = os.getenv("LOCAL_DEVELOPMENT", False)
+
+if LOCAL_DEVELOPMENT:
+    dotenv.read_dotenv(os.path.join(os.path.dirname(os.path.dirname(__file__)), '.env'))
```
ramonsaraiva: Why only local? Wouldn't it make sense to always use the `.env` file no matter what environment we're in?

apoclyps: I'm unfamiliar with the deployment process for this service however I typically use `.env` files for local development, and rely on configured environment variables on the instance hosting the application to provide the required values. 

@ramonsaraiva, I'm happy to make the change if there is a preference to always use the `.env` file.

ramonsaraiva: Not sure if there is a preference.
I usually use `.env` everywhere. Simply inject the proper `.env` for the proper environment.

apoclyps: I'll make the change this evening to use `.env` 👍

ramonsaraiva: Cool! Let's just make sure that we include the proper `.env` in the deployment process. I'm not sure who owns that but since we'll be relying on it, we need to make sure it is in the right place.

ramonsaraiva: Looking at the settings file, seems like we're already grabbing values from `os.environ` so we should be good.

apoclyps: Thanks for the feedback @ramonsaraiva. It's been actioned 👍please let me know if there is anything else I can do to help this PR get accepted :)

ramonsaraiva: I looks good on my side. I don't think `.env` will overwrite exported environment values, but if it does, whoever is responsible for the deployment will need to move the production exported values to `.env`.

apoclyps: From the [django-dotenv](https://github.com/jpadilla/django-dotenv) repository:
```
By default, variables that are already defined in the environment take precedence over those in your .env file. To change this, call read_dotenv(override=True).
```

For deploying to the existing environments, this change should have no impact.

apoclyps:
@olasitarska, @patjouk, and @aniav as the most active contributors to this repository do you have any guidance on who may be able to accept/reject this PR?

aniav:
@apoclyps I'm not sure anyone is doing current maintenance of this repository aside from small changes to the website appearance 🙌

amakarudze:
Sorry @apoclyps, this never got merged, there was a maintenance gap. Similar has been picked up by @marksweb. Thanks for your work on this!

PR #508, opened 2019-06-15
Add Persian translation to Coc

Rahimz:
Add Persian version of CoC.


magul:
We don't have anybody right now to do Persian review, so I'm going to merge that blindly (based on your contributions on Crowdin and discussion with @das-g on Slack).

Also asked @RachellCalhoun to create a Persian review group on GitHub and to put you as for the time being the only member of it.

PR #509, opened 2019-07-22
Run Django server from gulp

vanessa:
Adds the ability to run both `gulp watch` and the Django server in the same command. The good thing is that developers won't need to open 2 terminals to run static files tasks and the server.

Some checks to be made:
- [ ] Works on Ubuntu
- [ ] Works on Windows
- [ ] If the error message makes sense (the most common reason it would fail is because the virtualenv isn't activated, but I've added the stack nevertheless)

To reproduce, follow this branch's README and check if everything works properly


vanessa:
Hey @ramonsaraiva, do you know if there's anyone available to review this? 🙂

aniav:
I'm not sure we should be building any tooling around gulp to be honest, gulp is not widely used any more and I am not sure we should couple our codebase with it any more than it's coupled already. We should move to webpack if possible. What do you think?

vanessa:
@aniav I totally agree! I was thinking the use of gulp was still required since it's a tool relatively easy to configure. Do you want me to do this change when possible then?

aniav:
If you have time to introduce webpack instead of gulp that would be awesome ✨

PR #511, opened 2019-07-27
Upgrade website to Django 2.2

aniav:
Django 2.0 is a security vulnerability. We were not able to upgrade to newer versions because our virtualenv on PythonAnywhere was launched with Python 3.4 and upgrading it didn't run smoothly. It's now done so we can go on with the upgrade along with some other packages.


aniav:
Merged it as it didn't need review, just wanted CI to run tests against it :)

PR #513, opened 2019-09-17
Use case insensitive matching for user creation

### LINKED ISSUE #461: Organiser emails being treated as case sensitive
When organisers apply to host a workshop their emails seem to be being treated as case sensitive. This means that if we have a previous organiser using test@gmail.com and they apply to host another workshop using Test@gmail.com, then they have a second organiser account created for the new workshop.

seedhisadak:
fixes #461 

Use case insensitive matching for user creation. Don't create a new user if the same email with case insensitive match exists

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -268,7 +268,7 @@ class Event(models.Model):
             "is_active": True
         }
         user, created = User.objects.get_or_create(
-            email=email,
+            email__iexact=email,  # case insensitive match
             defaults=defaults
         )
         password = None


amakarudze (changes requested):
The code is untested and breaks 4 tests. Run `python -m pytest` in your local environment and fix the tests that are broken by your changes to ensure your change does not break anything. Thanks!

amakarudze:
Thanks for your input. There hasn't been any movement on this in over a year so now closing this.

PR #537, opened 2020-09-15
Add remote workshop option to application process

amakarudze:
Added a form before the application process to enable organizers to choose between Remote or In-Person workshop and provide information on the two types of workshops. 

To ensure we can keep the two post-Covid-19, changes have been done to both the Organize Form and EventApplication Model to enable storing the type of workshop in the database and data relevant to the type of workshop. The field ```remote``` will store the type of workshop in the model. All previous applications to be ```False``` by default.  

In-Person workshops have a field for ```safety``` added and Remote workshops will not require ```venue``` and ```safety``` information but will need details on ```tools``` used.


magul:
Two comments here @amakarudze (which are not connected with the content of pull request itself):
 * for few months we're serving tutorial from Github Pages; here's a commit that introduced this change: https://github.com/DjangoGirls/tutorial/commit/f7df5004acfc3e478c67f9cc775ca7ac96c91ff9
 * we have 19 languages there (no 8).

amakarudze:
> Two comments here @amakarudze (which are not connected with the content of pull request itself):
> 
> * for few months we're serving tutorial from Github Pages; here's a commit that introduced this change: [DjangoGirls/tutorial@f7df500](https://github.com/DjangoGirls/tutorial/commit/f7df5004acfc3e478c67f9cc775ca7ac96c91ff9)
> * we have 19 languages there (no 8).

Noted @magul. 
* I updated the number of languages. 
* I have maintained the original links to the tutorial. We are trying to migrate the Tutorial to the new GitBook app, any help would be appreciated.

magul:
@amakarudze 

> We are trying to migrate the Tutorial to the new GitBook app, any help would be appreciated.

Who is `we` here? Is there a dsicussion  in some public place? There was a discussion about migration from Legacy Gitbook (see: https://github.com/DjangoGirls/tutorial/issues/1652 closed in favour of https://github.com/DjangoGirls/tutorial/pull/1666).

amakarudze:
> @amakarudze
> 
> > We are trying to migrate the Tutorial to the new GitBook app, any help would be appreciated.
> 
> Who is `we` here? Is there a dsicussion in some public place? There was a discussion about migration from Legacy Gitbook (see: [DjangoGirls/tutorial#1652](https://github.com/DjangoGirls/tutorial/issues/1652) closed in favour of [DjangoGirls/tutorial#1666](https://github.com/DjangoGirls/tutorial/pull/1666)).

We is the foundation. Other resources have been moved to the GitBook app already.

magul:
> We is the foundation. Other resources have been moved to the GitBook app already.

Is there any publicly accesible channel that these discussions take place? There few people that could be really interested in it (cc: @aniav @das-g @@ekohl).

RachellCalhoun:
I checked this branch out locally and it lgtm :rocket:

aniav:
@magul I think there were a few things that happened independently. The tutorial got moved to github pages because of our discussions around plugins not working for the new gitbook version. Plugins that don't exist on the other gitbook repositories. While at the same time @amakarudze noticed gitbook legacy is going to go away and decided to move the books to the new gitbook version. So there are no bigger discussions happening, but I guess there is also no agreement on the level of engagement and ownership people take on different areas.

I think the main thing we can take from this situation is: if we make a bigger move around the repositories, just make sure to announce it on #committers on slack, we didn't do this for the main tutorial and it was not announced for the other books.  So @amakarudze had no idea what issues we were facing with upgrading the tutorial, @magul had no idea the other books got changed.

As for this PR, I'm updating our DNS settings so that the old domains point to the new versions of gitbooks. I'll let you know when this is done so we can change the URLs back 🙂

ekohl:
There indeed does appear to be a bit of a lack of communication. I'm always in favor of a consistent approach and would suggest to migrate everything to the approach we've taken in the tutorial. It doesn't rely on an additional service (GitBook) and only on Github. Since the repo is hosted there anyway, I think that's the best approach.

I am not on slack and don't intend to be. I'm also not following the other repos since I don't have commit rights there, but wouldn't mind keeping an eye on those as well.

Currently in beta, but there's also a discussions part @ Github. Perhaps that could also be interesting for a sort of meta/infra project for these kind of shared things. See https://github.blog/2020-05-06-new-from-satellite-2020-github-codespaces-github-discussions-securing-code-in-private-repositories-and-more/

magul:
I don't want to hijack this pull request so we can move that discussion somewhere else. I agree that we have some communication issues. I'm not sad, angry or dissapointed about that. I just believe, that we should maintain some kind communication channel (Slack probably, I'm there but haven't seen such a discussion) to talk such problems through.

I know, that @ekohl moved Tutorial from GitBook Legacy to Github Pages. I also believe we should have it consistent acroas the board, but as I mentioned above we should move that discussion somewhere else, as it's outside of the context of that pull request.

aniav:
I think the broader discussion is definitely something I'd vote on. I myself am only lurking here every now and then and sharing what I know from the past experiences, and I can give some guidance but I don't intend to be included in any broader changes 🙌 

For the current matter which is being able to process the remote workshops and have the links working properly on the new versions of the tutorial: I enabled DNS for the new gitbook on all the books aside from tutorial. We should get back to using the old links 🙌

ekohl:
:+1: for using the old links. Then we can migrate them later however we want.

core/validators.py
```diff
@@ -7,3 +9,11 @@ def validate_approximatedate(date):
             'Event date can\'t be a year only. '
             'Please, provide at least a month and a year.'
         )
+
+
+def validate_event_date(date):
+    today = datetime.today()
+    event_date = datetime.date(datetime.strptime('{0}-{1}-{2}'.format(date.year, date.month, date.day), '%Y-%m-%d'))
```
ekohl, line 16: This looks odd. Is this forcing a datetime or date object to a date? If it's already an instance of `date`, it shouldn't matter. If it is a datetime, `date.date()` would be the best way.

core/validators.py
```diff
@@ -7,3 +9,11 @@ def validate_approximatedate(date):
             'Event date can\'t be a year only. '
             'Please, provide at least a month and a year.'
         )
+
+
+def validate_event_date(date):
+    today = datetime.today()
```
ekohl, line 15: Later this is converted to a date. Why don't you use this:
```python
from datetime import date
today = date.today()
```

organize/forms.py
```diff
@@ -113,7 +117,16 @@ def clean_date(self):
         date = self.cleaned_data.get('date')
         validate_approximatedate(date)
         # TODO: add checking if the event is in the future
```
ekohl, line 119: Looks easy enough to implement in `validate_event_date` while you're at it. In fact, I think it should already be solved since this evaluates to `True`.
```python
date(2020, 10, 20) - date.today() < timedelta(days=90)                                                                                                                       
```

amakarudze:
@ekohl Thanks for your feedback. I incorporated your suggested changes. 

May you please review again with particular attention to this commit ```b5408a2``` in case you may be busy. I made some changes to the form wizard and models to open up both remote and in-person workshops and accept them based on merit. I would want them checked and verified before deploying these changes so I don't break the site in production. Thanks in advance!

ekohl:
Untested, but looks good to me.

PR #546, opened 2020-10-01
Improve Docs

Vipinkumar12it:


magul:
There's a typo here and from the @Vipinkumar12it 's pattern of behavior and because it's october today I would assume that is some kind of hacktober abuse.

I would suggest to close that one.

PR #548, opened 2020-10-01
Floating button for chat which will redirected to django girls gitter page

PremPawar:


amakarudze (changes requested):
Thank you for your work. Please take care of the following:
* The `<h1>Hello, World</h1>` is adding an extra heading and distorting the contents of the page. 
* The footer.html file should not have any static files in it and the changes you introduce with these static files breaks our page structure and conflicts with existing static files. Please remove them.
* May you please separate the 2 commits into 2 separate PRs so they can be reviewed separately?

amakarudze:
Thank you for your work on this. There hasn't been any progress in almost 10 months so we will be closing this. Thanks again.

PR #576, opened 2021-04-26
Add validation to EventApplication model

amakarudze:
* Add functionality to check if the organizer has an open application before submitting the application.
* Add functionality to check if the organizer has had an event in the past 6 months before submitting the application.
* Add tests to increase organize app coverage


RachellCalhoun (approved):
lgtm

PR #600, opened 2021-07-09
Replace TRAVIS with GITHUB_ACTIONS for STATICFILES_DIRS setting

amakarudze:


carltongibson (approved):
With this in play can you see if the tests pass removing the `cp -r` step in the workflow (that I added to get them working initially)? 

Looks good though.

amakarudze:
> With this in play can you see if the tests pass removing the `cp -r` step in the workflow (that I added to get them working initially)?
> 
> Looks good though.

Thanks @carltongibson. Let me do so.

PR #603, opened 2021-07-12
Check codecov integration with GitHub Actions

amakarudze:
Testing if codecov section works with GitHub Actions as part of GitHub Actions clean up.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -68,6 +68,5 @@ jobs:
         env:
           POSTGRES_PASSWORD: postgres
 
-# Can we restore this? Commented to get workflow running.
-#      - after_success:
-#          - bash <(curl -s https://codecov.io/bash)
+      - after_success:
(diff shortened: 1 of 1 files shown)


amakarudze:
@marksweb do you think you can help us integrate Codecov and GitHub Actions?

marksweb:
@amakarudze Yeah there's essentially a pre-built action for running coverage.

Is it codecov that you'd like?

https://github.com/marketplace/actions/codecov

amakarudze:
> @amakarudze Yeah there's essentially a pre-built action for running coverage.
> 
> Is it codecov that you'd like?
> 
> https://github.com/marketplace/actions/codecov

@marksweb We were using codecov before but we are no hard set on it I think. We just moved from Travis to Actions so I think we can use the built-in one. Thanks

amakarudze:
The issue was resolved by #642.

PR #604, opened 2021-07-12
Fix cache settings for GitHub Actions

amakarudze:
Remove cache settings copied from Travis file. They cause Github Actions to fail and seem unnecessary since builds pass without them.


carltongibson (commented):
Yep, OK. 

Worth reviewing https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows to see if we should apply that, but not urgent.

amakarudze:
Ok, will put this on hold for now. Thanks @carltongibson.

amakarudze:
@marksweb, could you help with this as well?

marksweb:
@amakarudze I can take a look 👍

amakarudze:
The issue was resolved by #643.

PR #605, opened 2021-07-12
Fix broken test in tests/organize/test_models.py

amakarudze:
This pull requests addresses issue [598](https://github.com/DjangoGirls/djangogirls/issues/598) and fixes the broken test.


carltongibson (approved):
So this looks good if you rebase after #606

PR #607, opened 2021-07-13
Fix date value error for event, revert to old settings

amakarudze:
Reverted to old comparison of approximate dates for event and removed the fixed false setting. Tests pass locally but this has been failing in production. Am using Python 3.7 in my local environment and the virtual environment in production is using 3.4. Not sure if this where the issue is stemming from.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -1,10 +1,10 @@
-from datetime import date
-
 import icalendar
 from django.conf import settings
 from django.http import HttpResponse, Http404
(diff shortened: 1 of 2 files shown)


carltongibson (commented):
> production is using 3.4. 

Can we nudge up there? If we can get to 3.7 then that would eliminate one variable.

amakarudze:
Let me merge this and do so now.

PR #616, opened 2021-08-09
Remove Python versions 3.5 and 3.6

amakarudze:
We no longer need new builds to pass for versions 3.5 and 3.6 since we have upgraded the production environment to Python 3.7 and would need to add later versions. This is also to prevent blocking PRs that pass for 3.7 because of versions we no longer support.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -33,8 +33,6 @@ jobs:
       max-parallel: 4
       matrix:
         python-version:
-        - 3.5
-        - 3.6
         - 3.7
 
     steps:


carltongibson (approved):
Looks good. When we’re ready we can add 3.8 in there, but I think other clean ups first for now. Thanks @amakarudze

marksweb:
Not sure if you've tried it yet, but tests won't run on 3.8 or later.

I think it was a package called `freezegun` which uses `time.clock` which doesn't exist from 3.8 onwards.

amakarudze:
> Not sure if you've tried it yet, but tests won't run on 3.8 or later.
> 
> I think it was a package called `freezegun` which uses `time.clock` which doesn't exist from 3.8 onwards. The project uses `0.3.11` but `1.1.0` is the latest release. It looks like `0.3.12` fixes this issue and they're now testing against 3.8, but not 3.9.
> 
> https://github.com/spulec/freezegun/blob/master/CHANGELOG

Yes, we have @marksweb . We tried up to 3.9 and after realizing we could only get tests to pass up to 3.7, we removed them and worked with 3.5, 3.6 and 3.7. We will work with upgrading packages gradually with 3.7 before moving to 3.8 or later.

PR #629, opened 2021-08-11
Stripe donations #572

marksweb:
Following on from my comment (https://github.com/DjangoGirls/djangogirls/issues/572#issuecomment-896412067) which detailed the initial implementation of this, I've now integrated it into the donation page and moved that page from `core` into the new `donations` app.

Screenshot 2021-08-11 at 14 59 59

There's some simple styling applied to the form that stripe injects via JS;

Screenshot 2021-08-11 at 15 10 01

Screenshot 2021-08-11 at 15 11 03

Some more could certainly be done to this, visually speaking to make it "nice", but frontend isn't my forte 😂 

I'll leave this as a draft for now & hopefully add some basic tests tonight/tomorrow.

One key thing here is the need for a stripe account & API keys. That'll need doing ahead of this going live - maybe I should build in some error catching 😬 

As a side note, I've been trying to use translation tags where I can as I'm making changes as I've seen the ticket talking about doing that and appreciate how big of a job that is.


amakarudze (changes requested):
Thanks @marksweb for all your work in this. Could you please remove donations/test.py since all our tests are in one module?

We are also going to need documentation for using .dotenv since this is not covered in our documentation. I take it, you have only done for the local environment like you mentioned earlier on and it won't break my production environment? 

Otherwise the Stripe work looks good. I am going to work on API keys for production so I can deploy this. Thanks again!

amakarudze (changes requested):
One thing I had missed is you have a `templates` folder in the `donations` app when all our templates are in the `templates` folder in the project root, so when I pulled your changes locally, clicking on the `Donate` link gives a 
```TemplateDoesNotExist at /donate/
donations/donate.html
```.

Could you move the templates for the `donations` app into the root `templates` folder? The folder structure should be `djangogirls/templates/donations`.

Thanks for all your work.

amakarudze:
> @amakarudze Good catch! I forgot about that `tests.py` that django created.
> 
> I've added a note about `python-dotenv` to the readme. It's a completely optional feature and without the file it won't do anything so you don't need to worry about production (if production runs `manage.py` commands).
> 
> Just to confirm, the two environment variables Stripe will need are `STRIPE_PUBLIC_KEY` and `STRIPE_SECRET_KEY`

Thanks @marksweb. I have the two variables now. Just need the folder structure for `donations` templates to be sorted out and I can merge this.

marksweb:
@amakarudze ok templates moved & even added some translation tags I missed before.

amakarudze (approved):

amakarudze:
Hey @marksweb, when I deployed this, it broke the site and had to revert the changes. When I tested locally, it didn't run because of the `.environment` file which was missing. I added this in production but the site wouldn't load. Could you separate the `dot-env` work from Stripe work so we can merge this? 

Thanks again for all your work. We really appreciate the help.

marksweb:
@amakarudze did you get a traceback for this?

I've had no issue running it without that file present so curious as to what happened.

And sorry about that 😬🤦‍♂️

amakarudze:
@marksweb no worries.

So there only traceback I can see in the error log on the server is for this issue [631](https://github.com/DjangoGirls/djangogirls/issues/631). I had to fix the slug manually after reverting your changes but it had already happened 2 times before. So not sure if the server error I got was because of this issue or because of `dotenv`. Maybe we can just do the PR for the stripe one separately for now. That way it's easy for me to separate issues when things break.

marksweb:
Ok thanks @amakarudze

It'd be good to know more about how the deployment is done incase it can be more robust. Or at least reproducible for testing.

I've got the stripe changes ready but spotted a gap in testing that I'm going to address first as I'd missed a link to the new donation page because the contribute page isn't tested.

amakarudze:
@marksweb We have a `deploy.sh` script we run which activates virtual environment, provides environment variables, fetches from repository, runs migrations, installs node packages and then reloads the website.

Thanks for adding tests for contribute page.

PR #637, opened 2021-08-16
Update dependencies with `pip-compile -U`

amakarudze:
After adding pip-tools to the project I added the command `pip-compile -U` to the deployment script, which updated requirements in production but not in dev. This pull request updates dependencies to match those in production.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -35,5 +35,4 @@ pytest-env==0.6.2
 pytest==3.9.1
 sentry-sdk==1.3.1
 slacker==0.9.65
-stripe
 vcrpy==2.0.1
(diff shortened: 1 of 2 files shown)


marksweb:
@amakarudze Just seen this. I wouldn't have the deployment script run `pip-compile -U`. This would cause the live site to run versions not yet used anywhere else.

I'd stick to that being a local/dev task. That way you can verify the versions it's going to upgrade and test them locally. Then commit the new `requirements.txt` and PR it like other changes.

Does that make sense?

amakarudze:
> @amakarudze Just seen this. I wouldn't have the deployment script run `pip-compile -U`. This would cause the live site to run versions not yet used anywhere else.
> 
> I'd stick to that being a local/dev task. That way you can verify the versions it's going to upgrade and test them locally. Then commit the new `requirements.txt` and PR it like other changes.
> 
> Does that make sense?

Yep, thanks. Let me close this.

PR #654, opened 2021-08-20
Fix for failing openstreetmap data

marksweb:
Some data being returned from open street map is different to that expected by a test, so this updates the expected data.

I've also disabled a test because the data from Patreon is in a different format to that expected by the test for the `fundraising_status` command.


marksweb:
Closing this because different results are had on here vs my local machine.

PR #658, opened 2021-08-25
Optimize applications page

### LINKED ISSUE #382: Optimize applications page.
Currently the applications page will create a DB query for each row in the applications list, to check if the request.user has already scored the application. This happens because the `scored_by_user` templatetag is using a db query. We should remove the query from the templatetag completelly. We should retrieve applications with their scores and anonymize them in python code while processing the list before passing it to the actual template and this way producing less queries.

vanessa:
This PR changes a few things:
* Optimize the applications scored_by_user query so we don't query if the user has scored a given application line by line
	* I added a test to make sure we're not doing any extra queries, but can remove if it's unnecessary
	* It's possible to optimize even further if we use annotations instead of properties in some places (e.g. `stdev`), paginate the applications, etc. but I didn't want to add more complexity to the PR, let me know if I should work on that
* Correctly anonymize sorting by `average_score` - now unscored applications will be at the end (added a test for that too)
* Move `get_applications_for_event` and `get_random_application` from `applications.questions` to `applications.services` (the `questions` module seemed to be more related to the form generation rather than anything else, but I don't have 100% context with the project so let me know if I should change it back)
* Add [ipdb](https://pypi.org/project/ipdb/) and [ipython](https://ipython.readthedocs.io/en/stable/) modules just for ease of local debugging/development (do we want to separate prod/dev requirements at some point?)
	* I think pip-tools ended up updating some of the libs when I ran `pip-compile -U`

### Sorting by average score

<table>
<tr>
	<td>Before
	<td>After
<tr>
<td>

https://user-images.githubusercontent.com/23136832/130745338-2e207b01-580a-40ad-85dc-6cb9a371ed3a.mp4
<td>

https://user-images.githubusercontent.com/23136832/130745159-4a4367a6-a14a-4b36-97e5-37208d8aa521.mp4
</table>

Tried to add comments to whatever I changed, but let me know if anything isn't clear enough, happy to improve if needed 🙂 

(Closes #382)

### DIFF
--- a/applications/services.py
+++ b/applications/services.py
@@ -19,9 +19,8 @@ def get_applications_for_event(event, state=None, rsvp_status=None, order=None, 
                           'form__event', 'scores__application')
     )
 
-    if user:
-        scores_subquery = Score.objects.filter(application=OuterRef("pk"), user=user)
-        applications = applications.annotate(was_scored_by_user=Exists(scores_subquery))
+    scores_subquery = Score.objects.filter(application=OuterRef("pk"), user=user)
+    applications = applications.annotate(was_scored_by_user=Exists(scores_subquery))
 
     if rsvp_status:
         applications = applications.filter(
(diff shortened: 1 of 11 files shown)


applications/services.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if order:
+        is_reversed = True if order[0] == '-' else False
+        order = order[1:] if order[0] == '-' else order
+        if order == 'average_score':
+            # here is an exception for the average_score, because we also want to get
+            # the standard deviation into account in this sorting
+            applications = sorted(
+                applications,
+                key=lambda app: (app.was_scored_by_user, getattr(app, order), -app.stdev()),
```
vanessa, line 40: 22, 23 and 40 were the lines that actually changed. sorry the diff got messed up :confused:

amakarudze (approved):

PR #661, opened 2021-09-05
Switch to django admin and upgrade to django 2.2

marksweb:
This change drops django-suit and upgrades to django 2.2

Although we'd said this set of changes should go into the dev branch, this change itself was actually really straightforward and I think it should to straight to main so that the site is running 2.2 rather than 2.0

I've also added django-debug-toolbar here to aid development.

Part of #628

### DIFF
--- a/templates/admin/base_site.html
+++ b/templates/admin/base_site.html
@@ -1,5 +1,5 @@
 {% extends 'admin/base.html' %}
-{% load staticfiles rev %}
+{% load rev %}
 
 {% block extrastyle %}
     {{ block.super }}
--- a/templates/admin/base_site.html
+++ b/templates/admin/base_site.html
@@ -1,7 +1,7 @@
 {% extends "admin/base.html" %}
 {% load i18n rev %}
 
-{% block title %}{{ title }} | {% trans "Django girls admin" %}{% endblock %}
+{% block title %}{{ title }} | {% trans "Django Girls admin" %}{% endblock %}
 
 {% block extrastyle %}
     {{ block.super }}
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block branding %}
-<h1 id="site-name"><a href="{% url 'admin:index' %}">{% trans "Django girls administration" %}</a></h1>
+<h1 id="site-name"><a href="{% url 'admin:index' %}">{% trans "Django Girls administration" %}</a></h1>
 {% endblock %}
 
 {% block nav-global %}{% endblock %}
--- a/templates/admin/base_site.html
+++ b/templates/admin/base_site.html
@@ -1,7 +1,7 @@
 {% extends "admin/base.html" %}
-{% load rev %}
+{% load i18n rev %}
 
-{% block title %}{{ title }} | {{ site_title|default:_('Django site admin') }}{% endblock %}
+{% block title %}{{ title }} | {% trans "Django girls admin" %}{% endblock %}
 
 {% block extrastyle %}
     {{ block.super }}
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block branding %}
-<h1 id="site-name"><a href="{% url 'admin:index' %}">{{ site_header|default:_('Django girls administration') }}</a></h1>
+<h1 id="site-name"><a href="{% url 'admin:index' %}">{% trans "Django girls administration" %}</a></h1>
 {% endblock %}
 
 {% block nav-global %}{% endblock %}
--- a/templates/admin/base_site.html
+++ b/templates/admin/base_site.html
@@ -1,7 +1,15 @@
-{% extends 'admin/base.html' %}
+{% extends "admin/base.html" %}
 {% load rev %}
+
+{% block title %}{{ title }} | {{ site_title|default:_('Django site admin') }}{% endblock %}
 
 {% block extrastyle %}
     {{ block.super }}
     <link rel="stylesheet" href="{% static_rev 'css/admin.css' %}">
 {% endblock %}
+
+{% block branding %}
+<h1 id="site-name"><a href="{% url 'admin:index' %}">{{ site_header|default:_('Django girls administration') }}</a></h1>
+{% endblock %}
+
+{% block nav-global %}{% endblock %}
--- a/templates/admin/organize/eventapplication/change_form.html
+++ b/templates/admin/organize/eventapplication/change_form.html
@@ -1,26 +1,32 @@
 {% extends "admin/change_form.html" %}
 {% load i18n %}
 
-{% block object-tools %}
-    <h4 class="italic-title">Triaging</h4>
-    <div class="box save-box">
-      {% if original.status != 'rejected' and original.status != 'accepted' %}
-        <a class="btn btn-high btn-info" href="{% url 'admin:organize_eventapplication_change_application_status' original.pk 'accepted' %}">
-          Accept application
-        </a>
-        <a class="btn btn-high" href="{% url 'admin:organize_eventapplication_change_application_status' original.pk 'in_review' %}">
-          Move to In Review
-        </a>
-        <a class="btn btn-high" href="{% url 'admin:organize_eventapplication_change_application_status' original.pk 'on_hold' %}">
-          Move to Hold
-        </a>
-        <a class="btn btn-high" href="{% url 'admin:organize_eventapplication_change_application_status' original.pk 'rejected' %}">
-          Reject application
-        </a>
-      {% else %}
-        Application closed.
-      {% endif %}
-    </div>
+{% block extrastyle %}
+{{ block.super }}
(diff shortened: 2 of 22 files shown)


carltongibson (commented):
Good work @marksweb — thanks for picking this up! 

Other than the look, are there any _functionality_ changes to take note of? 
(I'd say not to block on those, if there are any, but create smaller issues to re-add another way.) 

🎁

marksweb:
@carltongibson I'm not aware of any functionality changes. I've been through all admins and the only suit related thing I could really see was "form tabs". But that didn't seem to impact on that particular admin. Which is why I've raised this against main and not dev - it all looks like a safe & pain free upgrade.

The same can't be said for 3.2. I've made a start on that off the back of this change & that's where the work comes in to get things upgraded. Package upgrades and refs in the project to things like six, static have needed attention. But down from an initial test failure of 90 to 76 with those changes so I'll keep going on that one.

carltongibson:
@marksweb — Super. Getting back onto a supported Django is a big step. 💃

marksweb:
@carltongibson Indeed - it's ideal that 2.2 is still supported!

amakarudze:
> Good work @marksweb — thanks for picking this up!
> 
> Other than the look, are there any _functionality_ changes to take note of?
> (I'd say not to block on those, if there are any, but create smaller issues to re-add another way.)
> 
> 🎁

@carltongibson we define the look and feel of our admin interface using Suit in settings.py. 

@marksweb I will probably look at this PR tomorrow. Thanks for your work!

templates/admin/organize/eventapplication/change_form.html
```diff
@@ -1,26 +1,32 @@
 {% extends "admin/change_form.html" %}
 {% load i18n %}
 
+{% block extrastyle %}
+{{ block.super }}
+<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-KyZXEAg3QhqLMpG8r+8fhAXLRk2vvoC2f3B09zVXn8CA5QIVfZOJ3BCsw2P0p/We" crossorigin="anonymous">
+{% endblock %}
+
```
amakarudze, line 8: This page has different look from the rest of the admin pages. I am guessing it's missing some CSS style sheets. I have checked in both and Chrome and am getting the same result.

Screenshot 2021-09-07 at 18 44 52

Other pages look alright. See the screenshot below.
Screenshot 2021-09-07 at 18 44 23

amakarudze (changes requested):
The EventApplication Change Form is missing some CSS styling and is a bit distorted. 

Also, can we remove the Django Suit Copyright in the Admin footer since we are no longer using Django suit?

Otherwise, awesome work @marksweb!

templates/admin/organize/eventapplication/change_form.html
carltongibson: That doesn't look like the stock admin 🤔

marksweb:
Thanks @amakarudze - this set of changes (hopefully through to django 3.2) certainly needs a familiar eye to check things off.

I'll take a look at this later 👍

templates/admin/organize/eventapplication/change_form.html
marksweb: You're right @carltongibson - that is django-suit.

@amakarudze admin shouldn't look like that anymore because this change removes django-suit. Admin should look like this now;

Screenshot 2021-09-07 at 18 41 11

Specifically here, it's the application admin which I think is the only one that has extra functionality. It provides the "Triaging" form which uses bootstrap classes so I've also included bootstrap CSS in this specific admin template - bootstrap used to be made available through django-suit.

Ultimately I wouldn't worry about the visuals of the admin for this PR because it's only django 2.2. Django 3 changes the admin with the introduction of the sidebar, so reviewing the visuals once we're at 3.2 would be best.

Speaking of which, I've revisited things for that screenshot and been able to address the lack of branding. I was leaving this for the 3.2 PR.

Screenshot 2021-09-07 at 19 01 07

And on the subject of django 3.2 it's being held up by django-recaptcha. Tests are currently failing because it's not validating so I'm looking at options, including disabling recaptcha for tests.

templates/admin/base_site.html
```diff
@@ -1,7 +1,15 @@
-{% extends 'admin/base.html' %}
-{% load staticfiles rev %}
+{% extends "admin/base.html" %}
+{% load i18n rev %}
+
+{% block title %}{{ title }} | {% trans "Django girls admin" %}{% endblock %}
```
amakarudze, line 4: Can we make `g` a capital letter `G` in `girls`?

templates/admin/base_site.html
```diff
@@ -1,7 +1,15 @@
...
 
 {% block extrastyle %}
     {{ block.super }}
     <link rel="stylesheet" href="{% static_rev 'css/admin.css' %}">
 {% endblock %}
+
+{% block branding %}
+<h1 id="site-name"><a href="{% url 'admin:index' %}">{% trans "Django girls administration" %}</a></h1>
```
amakarudze, line 12: Same thing here.

amakarudze (changes requested):
Just need to capitalize Django Girls templates/admin/base_site.html.

templates/admin/organize/eventapplication/change_form.html
amakarudze: Ok, @marksweb, I needed to uninstall `django-suit` and `django-suit-redactor` to get the new interface. I guess I would need to uninstall them manually also in production.

I tried to migrate to Django 3.2 but couldn't get past the django-recaptcha glitch. Tests were failing so I just closed the PR.

marksweb: @amakarudze just run `pip-sync` after pulling or checking out to keep packages in sync.

Production should use this also to ensure old packages are removed.

Perhaps you can share the deploy script with me so I can advise?

And yeah, the django 3 move is more complicated than this one so it'll take me longer. I haven't raised a PR yet for it though, so not sure what you've closed.

templates/admin/base_site.html
marksweb: @amakarudze Done 👍

marksweb:
Just to reiterate from the comments here, I wouldn't deploy this until we can review the deploy script because we need to be sure that production runs `pip-sync` in order to clear out old packages when they're removed like they will be here.

amakarudze (approved):

amakarudze:
> Just to reiterate from the comments here, I wouldn't deploy this until we can review the deploy script because we need to be sure that production runs `pip-sync` in order to clear out old packages when they're removed like they will be here.

I have approved these changes but am holding off merging them and deploying them until the deploy issue is resolved. 

@marksweb the deploy script doesn't have `pip-sync` command for now, it just runs `pip install -r requirements.txt`.

marksweb:
@amakarudze Yeah that's what I imagined. What it'd be best doing is something like;

```
pip install --upgrade pip
pip install pip-tools
pip-sync
```

That way it's got the latest version of `pip`, it then installs `pip-tools` to ensure it's available and then the `pip-sync` will ensure only the packages we want exist.

amakarudze:
> @amakarudze Yeah that's what I imagined. What it'd be best doing is something like;
> 
> ```
> pip install --upgrade pip
> pip install pip-tools
> pip-sync
> ```
> 
> That way it's got the latest version of `pip`, it then installs `pip-tools` to ensure it's available and then the `pip-sync` will ensure only the packages we want exist.

Does this replace `pip install -r requirements.txt` or should I keep it in the deploy script @marksweb?

templates/admin/organize/eventapplication/change_form.html
amakarudze: Oh @marksweb  I meant I had done my own PR for migrating to Django 3.2 which I closed after django-recaptcha messed up things lol. 

The deploy script only has `pip install -r requirements.txt` for managing packages. I just need to know if the command is no longer required if we introduce the commands you introduced with `pip-sync` or it's still required.

marksweb:
> Does this replace `pip install -r requirements.txt` or should I keep it in the deploy script @marksweb?

Yes it replaces `pip install -r requirements.txt`

The `pip-sync` will do the same install from `requirements.txt`, but it'll also remove old packages no longer in `requirements.txt` and prevent any strange issues that could occur from having old packages lying around.

amakarudze:
> > Does this replace `pip install -r requirements.txt` or should I keep it in the deploy script @marksweb?
> 
> Yes it replaces `pip install -r requirements.txt`
> 
> The `pip-sync` will do the same install from `requirements.txt`, but it'll also remove old packages no longer in `requirements.txt` and prevent any strange issues that could occur from having old packages lying around.

Noted @marksweb. Merging this now and doing the deploy later tonight.

marksweb:
@amakarudze great. I should be around as I usually am if you need anything 👍

amakarudze:
Thanks @marksweb, this deployed with no issues. Still waiting to hear from PythonAnywhere about Postgres version upgrade. In the meantime, I just did a backup of the database just in case it gets wiped out during the upgrade.

carltongibson:
Super. Really good progress. Thanks @marksweb and @amakarudze

PR #665, opened 2021-09-09
Add Webucator logo to global partners

amakarudze:
We have a new global partner, Webucator. This pull request adds their logo to the list of global partners who already support us.


marksweb (approved):
Looks good

PR #670, opened 2021-09-13
Remove suit template tag

amakarudze:
There is a `suit_compat` that is breaking the `templates/admin/organize/view_change_status.html` template. This PR removes it.


marksweb (approved):
sorry, missed that!

PR #672, opened 2021-09-13
Removed email subject translations.

marksweb:
This removes translated email subjects to fix sentry error DJANGO-GIRLS-WEBSITE-E

Translation of emails would require a more complex process of storing a preferred language for a user and then making use of that in sending that user emails. So the translated stings here are pointless without that functionality.

### DIFF
--- a/organize/emails.py
+++ b/organize/emails.py
@@ -1,5 +1,4 @@
 from django.template.loader import render_to_string
-from django.utils.translation import gettext_lazy as _
 
 from core.emails import send_email
 
--- a/organize/emails.py
+++ b/organize/emails.py
@@ -6,7 +6,7 @@ from core.emails import send_email
 
 def send_application_confirmation(event_application):
     subject = (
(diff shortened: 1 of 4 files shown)


amakarudze:
Thanks @marksweb!

organize/emails.py
```diff
@@ -6,7 +6,7 @@
 
 def send_application_confirmation(event_application):
```
amakarudze, line 7: So this file has this unused import which is causing flake 8 to fail `from django.utils.translation import gettext_lazy as _`. Could you kindly remove it? Thanks @marksweb

amakarudze (changes requested):

organize/emails.py
marksweb: Ah yes, of course. Too busy rushing things. 🤦‍♂️

marksweb:
@amakarudze Fixed the unused import 👍

amakarudze (approved):

PR #699, opened 2021-09-23
Add code plugin to tinymce widget

### LINKED ISSUE #688: TinyMCE - Unable to switch to "Source Code" from the toolbar
image (1)
 I would like to edit the HTML code directly for our website contents (Django Girls Antananarivo workshop), but it doesn't have that option even if it says the code plugin is available.

ramonsaraiva:
Closes #688 

It was a bit tricky because of the way `django-tinymce` allows to modify the default settings..
If I just add the `code` plugin it overwrites the default `plugins` setting, same thing for `toolbar`, which results on an empty toolbar with the code widget.

Might not be the most beautiful implementation, but works as expected.

image
image

### DIFF
--- a/core/admin/forms/event_page_content.py
+++ b/core/admin/forms/event_page_content.py
@@ -1,12 +1,26 @@
+import copy
+
+import tinymce
+
 from django import forms
 from tinymce.widgets import AdminTinyMCE
+
+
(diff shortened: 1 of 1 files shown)


amakarudze (approved):
Thanks for your work on this @ramonsaraiva.

PR #708, opened 2021-11-24
Spanish Translations updated from POEditor

marksweb:


amakarudze (approved):

amakarudze:
@marksweb is there any work required to make these translations work on my end or merging and deploying this is all I need to do? This the first work we have received from PO Editor and I have no idea how the translations work. Thanks for your work on this.

marksweb:
@amakarudze Because the `.mo` file is present, there shouldn't be anything extra required. Sometimes you'd find only the `.po` file in version control and in that case you need `manage.py compilemessages` to create the `.mo` files.

So in theory this is good to go. The site doesn't have a language selector yet, so there shouldn't be any visible changes here. Visitors that have browsers set to Spanish may automatically get served Spanish content via the locale middleware.

Issue #625, opened 2021-08-10
Flickr integration to select more diverse photos


amakarudze:
When a new event is created, stock pictures already in the system are used as thumbnails for the event on the homepage as well as on the event homepage. However, these stock images are not as diverse and sometimes do not represent the event well. It would be great for the website to be integrated to the Django Girls Flickr account so that we have more diverse images to use.

Eloisa-R:
Hi!

I’m Eloisa, a frontend developer who attended a Django Girls event a few years back. I really enjoyed my time then and I wanted to contribute to the project, so I opened the PR #709 for this feature.

A couple of caveats:

1. While developing, I used my own Flickr API key, if you don’t have one you can request one here: https://www.flickr.com/services/apps/create/apply/. And then you will just need to set it as the value for the environment variable FLICKR_API_KEY.

2. I wasn’t sure about the criteria for selecting the photos, so I decided to call this method https://www.flickr.com/services/api/flickr.people.getPublicPhotos.html and then from the first page of results choose four random photos. I did filter out the photos that were in portrait format because I didn’t think they would fit well with the design of the page, but other than that it’s just random. I guess an improvement that could be made is getting photos from other pages as well, but I wasn’t sure if that was needed, so for now I left it like that.

3. I am more used to programming in Javascript than Python so I’m open to any feedback that you might have :)

Please let me know if there are any questions or comments!

PR #709, opened 2021-11-30
Flickr integration to select more diverse photos - Issue #625

Eloisa-R:
Implements the feature request from the issue #625 

As mentioned in the issue, there are a couple of things that could be useful to know:

1. While developing, I used my own Flickr API key, if you don’t have one you can request one here: https://www.flickr.com/services/apps/create/apply/. And then you will just need to set it as the value for the environment variable `FLICKR_API_KEY`.

2. I wasn’t sure about the criteria for selecting the photos, so I decided to call this method https://www.flickr.com/services/api/flickr.people.getPublicPhotos.html and then from the first page of results choose four random photos. I did filter out the photos that were in portrait format because I didn’t think they would fit well with the design of the page, but other than that it’s just random. I guess an improvement that could be made is getting photos from other pages as well, but I wasn’t sure if that was needed, so for now I left it like that.


marksweb (approved):

amakarudze (approved):
Great work @Eloisa-R! Thanks for picking this up.

Eloisa-R:
@amakarudze My pleasure! :)
By the way I saw that a pre-commit check failed when merging. Is it because of the linting errors? I didn't see them while developing locally, I'm sorry.
I can fix them in another PR if you want!

marksweb:
@Eloisa-R That's right. You should be able to see the output here; https://github.com/DjangoGirls/djangogirls/runs/4427049086?check_suite_focus=true

You can also install pre-commit to run this on your machine and be sure checks are passing prior to committing.

First do `pip install pre-commit` to ensure you have the package installed. Then run `pre-commit install` to install the hooks to your repo. Then because the files are already committed, run `pre-commit run --all-files` and you'll be able to fix the output.

If you could raise that in a new PR that'd be fantastic 👍

Eloisa-R:
Great, thanks for the help, @marksweb!
I'll fix it in a new PR :+1:

PR #711, opened 2021-12-06
Translations

amakarudze:
Add translations to Django Girls website.


amakarudze:
Seems these commits haven't been making it into the `main` branch. Just realised when I tried to deploy the changes from #708 that I had forgotten it was made against the `translations` branch.

marksweb:
@amakarudze my intention with the language support was to have it segregated so that it couldn't block main releases. For example if we got a lot of contributors helping with many languages it'd make it easier to sanity check the site on this separate translations branch before merging.

Hopefully that makes sense.

amakarudze:
> @amakarudze my intention with the language support was to have it segregated so that it couldn't block main releases. For example if we got a lot of contributors helping with many languages it'd make it easier to sanity check the site on this separate translations branch before merging.
> 
> Hopefully that makes sense.

Yes, it does actually. Have been busy with other things these past few months so I forgot to check this in a while. Will keep this in mind. Thanks for all your work in this @marksweb.

PR #719, opened 2022-01-04
Remove Oddbird from main sponsors

amakarudze:
Oddbird just cancelled their monthly sponsorship for us on Patreon and therefore we need to remove their logo from our list of sponsors.


marksweb (approved):
Looks good 👍

PR #721, opened 2022-01-25
Change support us to donate to us on menu

amakarudze:
Prospective donors are failing to find our donate page and the new stripe form because they visit the `Contribute` page instead of the `Support us` dropdown link to find the `Donate` link. This pull request renames the `Support us` link to `Donate to us` for easy visibility.


marksweb (approved):
Hopefully this has a positive impact 👍

PR #730, opened 2022-02-10
Update sentry-sdk

amakarudze:


marksweb (approved):
I've never had an issue updating sentry so no concerns here 👍

PR #746, opened 2022-03-06
Fix for events that have no date

marksweb:
If an event has no date the "is past" check fails on the event view. This fixes that.


amakarudze (approved):
LGTM!

PR #749, opened 2022-03-06
Pass stripe error in the session and improve success and error templates (#478)

marksweb:
This passes stripe errors through to the error template using the session so that the user gets a more detailed response to explain what went wrong.

This will resolve;
#478 
https://sentry.io/share/issue/bdb8bdbee6d24bd3bdd808b7f502b0fd/


amakarudze (approved):
LGTM!

PR #750, opened 2022-03-06
Remove distinct from the coaches queryset to allow deletion. [#739]

marksweb:
In #739 we can see that non-superusers are unable to delete coaches due to the `.distinct()` on their queryset.

As long as they should be able to delete coaches, this removes the `.distinct()` which should therefore allow the deletion by no longer throwing an error.


amakarudze (approved):
LGTM!

PR #751, opened 2022-03-07
Add new global partner Flatiron

amakarudze:
We have a new global partner sponsoring us through Patreon. This PR adds their logo to our main sponsors section.


marksweb (approved):
Let's get this out 👍

PR #753, opened 2022-03-09
Update Patreon figures

amakarudze:
We have another supporter on Patreon and have also changed the goal from $1,500 to $2,500. This PR updates the fundraising status on Patreon.


marksweb (approved):
That's great news!

PR #761, opened 2022-03-30
Fix triaging options for a deployed event [#758]

marksweb:
Correct me if I'm wrong, but based on the error in #758 the triaging options shouldn't be available to deployed events.

This change will hide the triaging buttons for a deployed event and instead display "Application deployed."

### DIFF
--- a/organize/admin.py
+++ b/organize/admin.py
@@ -4,7 +4,7 @@ from django.urls import path
 from django.utils.translation import gettext_lazy as _
 
 from .models import EventApplication, Coorganizer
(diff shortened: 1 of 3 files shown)


amakarudze:
Thanks for this @marksweb. 

I am verifying our process. There are times we may need to cancel an event because organizers have violated our terms and conditions. Am checking with @GirlGeekUpNorth about how we would proceed on that. Do we need to reject a deployed event and then delete/disable it or what? Once that is cleared, I can then make an appropriate review.

Thanks again for all your work on this repo!

amakarudze (approved):
Looks good to me! Thank you for your work on this @marksweb!

PR #763, opened 2022-03-31
Duplicate event for events with English and non-English websites

amakarudze:
This PR enables the Awesomeness Ambassador to duplicate event for events that have English and non-English websites by cloning the event.

### DIFF
--- a/core/models/event.py
+++ b/core/models/event.py
@@ -288,7 +288,7 @@ class Event(models.Model):
         clone = Event.objects.get(id=self.id)
         clone.id = None
         clone.name += ' (clone)'
-        clone.page_url += '_(clone)'
+        clone.page_url += '_clone'
         clone.save()
         return clone
 
--- a/core/models/event.py
+++ b/core/models/event.py
@@ -287,7 +287,7 @@ class Event(models.Model):
         # the database as a new record by wiping the ID and re-saving
         clone = Event.objects.get(id=self.id)
         clone.id = None
-        clone.name += ' (clone)'
+        clone.name += ' clone'
         clone.page_url += '_clone'
         clone.save()
         return clone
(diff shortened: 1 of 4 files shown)


core/models/event.py
```diff
@@ -272,6 +272,26 @@ def set_random_cover(self):
...
+        if not self.id:
+            # this Event isn't saved yet...
+            return None
+
+        # first, get a copy of self (that isn't self), we'll save this into
+        # the database as a new record by wiping the ID and re-saving
+        clone = Event.objects.get(id=self.id)
+        clone.id = None
+        clone.name += ' (clone)'
+        clone.page_url += '_(clone)'
```
marksweb, line 291: @amakarudze if this is a url we probably don't want brackets in as they're not valid characters in a url.

templates/admin/base_site.html
```diff
@@ -1,15 +1,17 @@
 {% extends "admin/base.html" %}
 {% load i18n rev %}
+{% load static %}
```
marksweb, line 3: Minor thing but it's a bit neater to load static in the line above & avoid using load twice.

marksweb (changes requested):

templates/admin/base_site.html
amakarudze: Noted. Done

marksweb: Sorry! I meant, `{% load i18n rev static %}` but was reviewing from my phone so didn't include an example. So only load in what's needed in 1 line, rather than static separately.

amakarudze:
Hi @marksweb. May you kindly check the changes you requested before I can merge this? Thanks

marksweb (approved):

marksweb:
@amakarudze yeah this looks good. 👍

amakarudze:
> @amakarudze yeah this looks good. 👍

Thanks @marksweb 🥳

PR #769, opened 2022-04-18
Fix the homepage link to Sentry.

bhushan-mohanraj:
From the logo, I believe the link was intended to point to https://sentry.io/welcome/ rather than https://www.sentry.com/.


amakarudze (approved):
Thanks for your work on this @bhushan-mohanraj! ✨

PR #772, opened 2022-04-27
Add robots.txt and sitemap.xml routes on the website

amakarudze:
This PR adds a robots.txt and sitemap.xml to the website to enable our website to be crawled by robots and indexed by search engines such as Google. This is the first step in addressing issue #766. Would need to submit the website's sitemap to [Google](https://support.google.com/webmasters/answer/7451001?hl=en#zippy=%2Csubmit-a-sitemap) to enable our blog posts to be indexed and create more visibility for our blog and sponsors.

PR #773, opened 2022-04-28
Remove `/` in the sitemap url

amakarudze:
Just got a warning on the server about a trailing `/` after deploying #772 so this pull request addresses the warning.


marksweb (approved):

marksweb:
@amakarudze confirmed this is correct for a path 👍

PR #790, opened 2022-05-29
fix: Change the use of city to page_url in the applications app [#682]

### LINKED ISSUE #682: Applications app uses `city` in it's URLs and for referring to events but should use `page_url`
Recently the core URLs for event were updated to use `page_url` instead of `city` as their `Event` attribute, because `city` isn't a unique value, or valid slug. Instead the `Event` object has a `page_url` field which is the slug value created when an event application is deployed.

The applications app still uses `city` in it's URLs and in the views this `city` variable is used to get `Event` objects. This needs updating to use `page_url` in the same way that the core app was in order to ensure the correct objects are being retrieved from the database.

marksweb:
This fixes #682 to move the use of the `city` value to `page_url` for events as `page_url` is used in the URL patterns.


marksweb:
@amakarudze I _think_ this covers #682 but please let me know if I've missed anything.

amakarudze (approved):
LGTM @marksweb. Thank you for your work on this ⚡️✨🥳!

Issue #767, opened 2022-04-11
Prohibit links in contact form to limit spam


amakarudze:
We have started receiving spam messages from the hello@djangogirls.org email address from our website. There is a need to prevent messages with links from submitting as well as enforce DMARC rejection and quarantine policies to prevent our domain from being misused.

amakarudze:
Seems the message did not come through our contact form or website. So this is a DMARC issue.

marksweb:
If we've got access to a django form for this, then we can implement django-bleach to control what tags can be added (potentially).

amakarudze:
Hey @marksweb, we do have access to the django form. Let me look at django-bleach. Thanks!

PR #793, opened 2022-06-21
Add BleachField to contact form and model

amakarudze:
Addresses #767.


djangogirls/settings.py
```diff
@@ -274,3 +275,15 @@ def gettext(s):
 RECAPTCHA_TESTING = os.environ.get("RECAPTCHA_TESTING") == 'True'
 if RECAPTCHA_TESTING:
     SILENCED_SYSTEM_CHECKS = ['captcha.recaptcha_test_key_error']
+
+BLEACH_ALLOWED_TAGS = ['p', 'b', 'i', 'u']
+
+BLEACH_ALLOWED_ATTRIBUTES = ['title']
+
+BLEACH_ALLOWED_PROTOCOLS = [
```
marksweb, line 283: I don't think this setting is needed because links aren't allowed.

marksweb (changes requested):
I've added a few comments.

djangogirls/settings.py
amakarudze: Ok, will remove this as well.

ramonsaraiva:
Do we still want to bleach the message field? I believe it is only missing adding that unit test check, right?

marksweb:
@ramonsaraiva yeah I think it's still a worthwhile addition.

ramonsaraiva:
@marksweb I added changes to the comments you've made here https://github.com/DjangoGirls/djangogirls/pull/834 - closing this draft.

PR #804, opened 2022-09-30
Attempt to fix #799 (Unique violation on application form)

### LINKED ISSUE #799: IntegrityError: duplicate key value violates unique constraint "applications_application_form_id_b47c846a_uniq"
Sentry Issue: [DJANGO-GIRLS-WEBSITE-43](https://sentry.io/organizations/django-girls/issues/3486070965/?referrer=github_integration)

```
UniqueViolation: duplicate key value violates unique constraint "applications_application_form_id_b47c846a_uniq"
DETAIL:  Key (form_id, email)=(675, ) already exists.

  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)

IntegrityError: duplicate key value violates unique constraint "applications_application_form_id_b47c846a_uniq"
DETAIL:  Key (form_id, email)=(675, ) already exists.

(12 additional frame(s) were not displayed)
...
  File "django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
```

ramonsaraiva:
This PR addresses what I mentioned on https://github.com/DjangoGirls/djangogirls/issues/799#issuecomment-1262565181 - I'm not sure if this is the *perfect* fix but can potentially avoid those unique violations in the future. It also avoids saving the application multiple times during the same form save.

### DIFF
--- a/applications/forms.py
+++ b/applications/forms.py
@@ -4,6 +4,7 @@ import requests
 from django import forms
 from django.conf import settings
 from django.core.mail import EmailMessage
+from django.db import transaction
 from django.template.loader import render_to_string
 from django.utils.translation import gettext_lazy as _
 
@@ -52,7 +53,8 @@ class ApplicationForm(forms.Form):
         return cleaned_data
 
     def save(self, *args, **kwargs):
-        application = Application.objects.create(form=self.form)
+        application = Application(form=self.form)
+        answers = []
 
         for name in self.cleaned_data:
             question = None
@@ -66,20 +68,14 @@ class ApplicationForm(forms.Form):
                         application.newsletter_optin = True
                     else:
                         application.newsletter_optin = False
-                    application.save()
 
             value = ', '.join(value) if type(value) == list else value
 
             if question:
-                Answer.objects.create(
-                    application=application,
-                    question=question,
-                    answer=value
-                )
+                answers.append(Answer(question=question, answer=value))
 
                 if question.question_type == 'email':
                     application.email = value
-                    application.save()
 
         if not self.form.event.email:
             # If event doesn't have an email (legacy events), create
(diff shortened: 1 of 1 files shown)


ramonsaraiva:
@marksweb @marksweb thoughts?

marksweb:
@ramonsaraiva i think you're right.

This should avoid it.

I alao think the unique constraint should be removed. If email can be empty then it shouldnt be a constraint

ramonsaraiva:
I can agree with that too, we can also make the unique constraint only apply when e-mail is not null. Something similar to:
```python
models.UniqueConstraint(
    fields=["form", "email"],
    condition=models.Q(email__isnull=False),
    ...
)
```

marksweb:
@ramonsaraiva yes good point, that's a real benefit of those new constraints.

What Django version did they get introduced? I think I upgraded the project to 3.2.

ramonsaraiva:
@marksweb I think that `UniqueConstraint` fields and conditions are available for a long time.. At least 2.2

marksweb:
@ramonsaraiva You're right, it's just been extended in 4.0 - I knew there was something new recently...

https://docs.djangoproject.com/en/4.0/ref/models/constraints/#expressions

ramonsaraiva:
so I guess I can also add this unique constraint as part of these changes?

marksweb:
@ramonsaraiva yeah that would be great!

amakarudze:
@ramonsaraiva is this PR done?

ramonsaraiva:
I'll just add the unique constraint as part of it and then it should be ready to go.

ramonsaraiva:
@marksweb @amakarudze should be ready right now.

amakarudze (approved):

PR #810, opened 2022-12-12
Classify sponsors

amakarudze:
We have just been adding sponsor logos randoms. To attract new sponsors in the next year, have decided to classify sponsors by how much they are contributing per year, listing the major contributors per year.


templates/includes/_main_sponsors.html
```diff
@@ -26,6 +52,19 @@
                  alt="Sherpany" style="margin-top: 50px" />
         </a>
     </div>
+    <div class="col-md-3">
+        <a href="https://runcode.io/">
+            <img class="img-responsive center-block" src="{% static 'img/global/supporters/runcode.png' %}"
```
marksweb, line 57: @amakarudze It looks like this image is missing

amakarudze: Have added the logo, thanks for catching that @marksweb.

marksweb (approved):

PR #816, opened 2022-12-20
Add Diamond sponsor level

amakarudze:
Am planning to make a change to the sponsor levels for 2023, replacing Bronze with Diamond.

### DIFF
--- a/templates/includes/_main_sponsors.html
+++ b/templates/includes/_main_sponsors.html
@@ -3,13 +3,13 @@
 
 <div class="row">
     <div class="col-md-4"></div>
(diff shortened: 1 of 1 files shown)


marksweb:
Just a minor thing on the sponsors - I think the level titles could have a bigger font or a heavier font weight - they don't really stand out like titles should.

marksweb (approved):

amakarudze:
> Just a minor thing on the sponsors - I think the level titles could have a bigger font or a heavier font weight - they don't really stand out like titles should.

Let me make those changes, thanks @marksweb. Was hoping to use some badges later, if we get a good designer.

PR #827, opened 2022-12-30
Update Slack integration to use SDK

### LINKED ISSUE #787: Move from using archived `slacker` API to Slack's own API to integrate website and Slack
We are still using [slacker](https://github.com/os/slacker) API to integrate the website despite the API being archived in 2020 and Slack introducing their own [API](https://api.slack.com/). Since [slacker](https://github.com/os/slacker) is no longer maintained, we need to update our website to use [Slack's own API](https://github.com/slackapi/python-slack-sdk).

vanessa:
## Changes
* Remove deprecated lib `slacker`
* Replace calls to Slack API with their own SDK
* Pin pip-tools (it's always good to pin requirement versions to help prevent compatibility issues and unexpected behavior in production)
* Rename `SLACK_API_KEY` to `SLACK_BOT_TOKEN` for explicitness

## Notes
* To send the message with a custom username and icon_emoji, the app needs the [chat.write:customize](https://api.slack.com/scopes/chat:write.customize) scope (I think this is new, so someone might need to update the permissions on the Slack dashboard)
* It's now required to pass a team and channel IDs when inviting users to Slack (in `user_invite_to_slack`), so some environment variables were added: `SLACK_TEAM_ID`, `SLACK_INVITE_CHANNEL_IDS` (the latter is a comma-separated list with channel IDs) - [here's a guide on how to find these IDs](https://feedly.helpscoutdocs.com/article/648-how-to-find-slack-channel-id)
* I couldn't test the `user_invite_to_slack`, tried to test locally but it seems the scope required to invite users ([admin.users:write](https://api.slack.com/scopes/admin.users:write)) is limited to workspaces in Enterprise Grid

## Actions Required

These are the actions required from maintainers to make sure the new code works as expected.

- [ ] Set `SLACK_TEAM_ID`, `SLACK_INVITE_CHANNEL_IDS` settings on production
- [ ] Make sure the app has a chat.write:customize scope on Slack
- [ ] Rename `SLACK_API_KEY` to `SLACK_BOT_TOKEN` on production
- [ ] After PR is deployed, test `user_invite_to_slack`

This PR should fix #787

### DIFF
--- a/tests/core/test_commands.py
+++ b/tests/core/test_commands.py
@@ -54,10 +54,6 @@ def test_add_organizer(click_runner, future_event):
 
 def test_new_event_with_one_organizer(click_runner, random_day, events, slack_mock, settings):
     settings.ENABLE_SLACK_NOTIFICATIONS = True
-<<<<<<< HEAD
-=======
-
->>>>>>> 2d9c7a6 (Enable Slack mock for testing)
     assert Event.objects.count() == 4
 
     command_input = (
--- a/tests/core/test_commands.py
+++ b/tests/core/test_commands.py
@@ -54,6 +54,10 @@ def test_add_organizer(click_runner, future_event):
 
 def test_new_event_with_one_organizer(click_runner, random_day, events, slack_mock, settings):
     settings.ENABLE_SLACK_NOTIFICATIONS = True
+<<<<<<< HEAD
+=======
+
+>>>>>>> 2d9c7a6 (Enable Slack mock for testing)
     assert Event.objects.count() == 4
 
     command_input = (
--- a/tests/core/test_commands.py
+++ b/tests/core/test_commands.py
@@ -1,6 +1,5 @@
 import random
 from datetime import date
-from unittest import mock
 
 import pytest
 import vcr
@@ -43,8 +42,7 @@ def test_update_coordinates(click_runner, past_event):
     assert past_event.latlng == latlng
 
 
-@mock.patch("core.models.User.invite_to_slack")
-def test_add_organizer(_, click_runner, future_event):
+def test_add_organizer(click_runner, future_event):
     assert future_event.team.count() == 1
 
     command_input = f"{future_event.pk}\n" "Jan Kowalski\n" "jan@kowalski.example.org\n" "N\n"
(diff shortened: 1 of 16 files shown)


tests/mocks/slack.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from easydict import EasyDict
+
+from core.slack_client import slack_client
+
+
+@pytest.fixture()
+def slack_mock(mocker):
```
vanessa, line 8: I was trying to mock Slack calls for testing and found the VCR approach to be a bit confusing. I had problems because it was trying to compare the host, headers, and the exact URL path, which is already tested by the Slack library itself. I figured we could benefit from a simpler way to mock and would love to hear your thoughts on this.

tests/core/test_commands.py
```diff
@@ -101,10 +102,12 @@ def test_new_event_with_two_organizers(click_runner, random_day, events):
     assert Event.objects.count() == 5
     event = Event.objects.order_by('pk').last()
     assert event.team.count() == 2
+    slack_mock.chat_postMessage.assert_called_once()
```
vanessa, line 105: This is an example of the mock in use.

tests/mocks/slack.py
ramonsaraiva: Maybe you can do the same thing I did on https://github.com/DjangoGirls/djangogirls/pull/833 - deleting the existing cassette files that the tests are using and regenerating them by running the tests again.

vanessa: I had a couple of issues that are worth writing down for future reference if needed:
* As to posting messages, I wanted to make sure the request to Slack was sent and went through, but Slack requires me to pass an authorization header with the API token, otherwise the response will fail, and we can't commit the token in cassette files (it was automatically being added to it).
* I couldn't test `user_invite_to_slack` because I don't have the right test environment (workspace in Enterprise Grid) as DG has to call the app with this scope, so the cassette would have failed response which I don't think makes much sense.

Overall, the gist of the issue seems to be that I'm trying to test internal stuff from DG's workspace, which I don't have access to, and I'm just not familiar with vcrpy. I guess [filtering some fields from the request](https://github.com/kevin1024/vcrpy/blob/964615af2562b24af7864456bb4c347956def424/docs/advanced.rst#filter-sensitive-data-from-the-request) should solve the issues.

amakarudze: Hi @vanessa. Thank you for your work on this. We are currently on the Pro Plan and Slack gives discounts for non-profits on Pro and Business Level. Let me see if I can reach out about an upgrade so we can get on the Enterprise Grid but I have to run this by the trustees first. Will let you know once that is in order but for now I will just approve this and see what I can configure in the current set up.

vanessa: @amakarudze do we want to merge this PR and test on production?

amakarudze: Yes we do @vanessa but I have been working on getting a Slack upgrade so we can be able to test it out. Right now our Pro account cannot support this change. That is the only hold-up to this. Maybe you can fix the conflicts and we merge this in while we wait?

PR #830, opened 2022-12-31
Enable use of Docker for local deployment

### LINKED ISSUE #821: Enable use of Docker for local deployment
The current set up of the website involves many manual steps and hence it takes up a new contributor a few hours to set up the website locally. Introducing Docker will make setting up the project locally easier.

ramonsaraiva:
Closes #821 

Draft with simple docker and docker compose setup for making local environments simpler for new (and existing) contributors.

This is a pretty early stage attempt that runs the app and a database container and spins up a local server exposing the `8000` port.

A few points that still need to be addressed, but not sure what would be our approach to them:

- Instructions to watch static files (with gulp watch)
- Instructions to run tests inside the app container

Let me know if yall think we need more functionality on this initial setup or if we should create followup issues for them.. Keeping as a draft PR for now.


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    && npm install -g gulp
+
+COPY rootfs /
+COPY . ${APP_DIR}
+
+RUN gulp local
+
+ENTRYPOINT ["/entrypoint.sh"]
+
+HEALTHCHECK \
```
vanessa, line 29: oh this is nice! didn't know about it

marksweb: You can run scripts for health checks as well @vanessa 

I've got one to check the available memory to kill off containers that run out of RAM.

```
HEALTHCHECK --interval=12s --timeout=12s --start-period=30s \
    CMD ./check-memory.sh
```

Where that script does
```
#!/bin/bash
musage=$(free | awk '/Mem/{printf("RAM Usage: %.2f%\n"), $3/$2*100}' |  awk '{print $3}' | cut -d"." -f1)
if [ $musage -ge 55 ]; then
echo "Current Memory Usage: $musage%"
exit 1
else
echo "Memory usage is in under threshold"
exit 0
fi
```

core/views.py
```diff
@@ -174,3 +174,7 @@ def coc_legacy(request, lang=None):
 #         'donor_list': donor_list,
 #         'quotes': DONOR_QUOTES,
 #     })
+
+
+def healthcheck(request):
```
marksweb, line 179: I wouldn't do this. The health check should check that the home page loads because that's the application we want to be healthy.

This view may return what you want while something in the application is broken and preventing the home page from rendering. Does that make sense?

docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      - POSTGRES_PASSWORD=djangogirls
+      - POSTGRES_HOST=db
+      - POSTGRES_PORT=5432
+      - POSTGRES_DB=djangogirls
+    depends_on:
+      - db
+    volumes:
+      - .:/var/www/app
+    ports:
+      - "8000:8000"
```
marksweb, line 22: Where ports are defined in compose, I find it really useful to allow for port overrides.

Quite often I've got a number of projects running, so I need to set different ports.

For example, on my db container, I specify;

```
  db:
    image: postgres:12.5-alpine
    environment:
      POSTGRES_USER: docker
      POSTGRES_PASSWORD: docker
      POSTGRES_DB: docker
    ports:
      - "${DB_PORT:-5432}:5432"
```

Then I can say `DB_PORT=5433 docker compose up`

And for the app, I'd typically use `- "${HTTP_PORT:-8000}:8000"`

marksweb (changes requested):

core/views.py
ramonsaraiva: Yeah that makes sense, i'll change it to hit the homepage in healthcheck.sh and remove this view.

ramonsaraiva:
Are we getting this one in? @amakarudze @marksweb

amakarudze:
> Are we getting this one in? @amakarudze @marksweb

Yes we are but only when it's no longer marked as a draft and conflicts are resolved @ramonsaraiva.

ramonsaraiva:
Awesome, fixed the conflict and set as ready for review.

README.md
```diff
@@ -74,7 +74,14 @@ Step into newly created `djangogirls` directory:
 cd djangogirls
 ```
 
+### Docker
+
+If you have Docker and Docker compose installed, run `docker-compose up`
```
marksweb, line 79: Compose is now part of docker. So now you can run `docker compose`

marksweb (approved):

amakarudze (approved):

PR #833, opened 2023-01-03
Calling flickr API when loading modules

### LINKED ISSUE #805: Calling flickr API when loading modules
Today my connection was a bit unstable when I was setting up the repo in a different device and `./manage.py createsuperuser` failed.

I was trying to understand why and then I realized we're calling `select_random_flickr_photos(4)` in the module scope (`core/default_eventpage_content`). That module is imported in one of the models files (`core/models/event` I think) and it gets called whenever Django is initialized.

I think we need to make that lazy so it doesn't get called when its not gonna be used.

ramonsaraiva:
Closes #805

### DIFF
--- a/core/default_eventpage_content.py
+++ b/core/default_eventpage_content.py
@@ -1,6 +1,8 @@
 # -*- encoding: utf-8 -*-
 import random
 
+from functools import lru_cache
+
 from django.conf import settings
 from django.core.files.uploadedfile import UploadedFile
 from django.utils.translation import gettext_lazy as _
@@ -25,9 +27,6 @@ DEFAULT_BACKGROUND_PHOTOS = {
 }
 
 
-four_random_flickr_photos = select_random_flickr_photos(4)
-
-
 def get_random_photo(section):
     if section in DEFAULT_BACKGROUND_PHOTOS:
         photos = DEFAULT_BACKGROUND_PHOTOS[section]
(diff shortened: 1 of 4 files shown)


ramonsaraiva:
Hmm, those test failures seem related to vcrpy not being able to overwrite the yaml file with the new request being made to flickr with the new changes (before the first request to flickr was done during the module intiialization).

What process have we been using for this? Do I regenerate the vcr yaml output locally? (In this case, most requests will probably have API key failure responses as I don't have those in my local environment).

vanessa:
I had a similar issue when trying to update the Slack-related stuff https://github.com/DjangoGirls/djangogirls/pull/827#discussion_r1059705243

amakarudze:
> Hmm, those test failures seem related to vcrpy not being able to overwrite the yaml file with the new request being made to flickr with the new changes (before the first request to flickr was done during the module intiialization).
> 
> What process have we been using for this? Do I regenerate the vcr yaml output locally? (In this case, most requests will probably have API key failure responses as I don't have those in my local environment).

Hey @ramonsaraiva, I think I have had to regenerate vcrpy locally in the past once when it was necessary to do so to get the tests to pass. We don't have API keys in dev and CI environment but tests still pass so regenerating the vcrpy yaml locally and running the tests should fix it. 

I haven't had time to look at your code work but the initial flickr integration was done here #709. Will try to look at it tomorrow.

ramonsaraiva:
@vanessa Yeah, this seems to be a similar problem.

@amakarudze I was able to regenerate the vcr files locally and tests passed, but the output of the response payloads are things like "not authorized" or "missing authentication token" etc, different from the existing responses which seem to be a successful response from each API.

I don't think that's a big problem, but ideally we would have real requests/responses in the cassette files so the mocks are more realistic.

I guess that's the way to go tho.

ramonsaraiva:
@vanessa @amakarudze generated the new cassette files locally, now tests pass :-) But as you can see, the headers/payloads are quite different from real and successful requests/responses to those APIs. I guess that's fine tho. LMK if you think I need to do anything else.

PR #841, opened 2023-01-17
ci: Add dependabot config to keep actions updated

marksweb:
Using this config, dependabot will keep the versions within github actions (things like checkout, setup python etc) updated. It's set on a monthly basis which I find works well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,13 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "monthly"
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+
+  - package-ecosystem: "pip-compile"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


ramonsaraiva:
Wouldn't this stop the `pip` package ecosystem? Or that's enabled by default? I would believe that if we're "overwriting" or adding a custom dependabot configuration we'd need to set the pip package ecosystem as well. I guess even if it adds by default, we could be explicit about it and add that ecosystem with the interval that we want anyways. Wdyt?

marksweb:
@ramonsaraiva This is completely unrelated to `pip` and the project itself.

I've got this same file in django-sql-explorer. See here; https://github.com/groveco/django-sql-explorer/blob/master/.github/dependabot.yml

What it does is it gets the dependabot bot to raise PRs to upgrade action versions. See here;
https://github.com/groveco/django-sql-explorer/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+

ramonsaraiva:
Hmm, yeah I think I understand the part that this gets the dependabot to upgrade action versions.. But in the example you've shown in `django-sql-explorer`, I don't see dependabot PRs for the requirements files, but maybe that's intentional.

My only doubt is if whenever we add this `dependabot.yml` file, the bot will stop automatically creating PRs to upgrade the deps versions in our requirements file, because you're not adding the `pip` package ecosystem in it.

i.e.:
```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"

  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "weekly"
```

marksweb:
@ramonsaraiva Yeah I don't get dependabot to do `pip` packages. That's better kept to the development workflow.

If there's a node pipeline, it's good to get dependabot to handle node as well.

amakarudze (changes requested):
Thanks @marksweb but like @ramonsaraiva I kind of appreciate `pip` and `node js` dependabot workflows in keeping packages up-to-date so can we add them to this workflow?

marksweb:
@amakarudze Yeah sure 👍

PR #842, opened 2023-01-17
ci: Add django-upgrade and python-upgrade pre-commit hooks

marksweb:
This looks like rather a big review, because there's a lot of files changing. But it's mostly automated tools, and this is preparation for future upgrades of python and django. This first adds the pre-commit hooks to upgrade python and django specific code to a target version which is currently set at python 3.10 and django 3.2. This will ensure that any new code committed will be as efficient as possible using these versions of python & django.

What these hooks also do is enable future upgrades to happen really quickly. You can just update the target version, for example to django 4.1. Then run `pre-commit run --all-files` and it will update the project for django 4.1.

Alongside these hooks in `pre-commit-config.yml`, I've added the CI config, because adding the pre-commit ci tool to the repo is a really good way to verify people raising PRs have actually setup pre-commit locally. It'll run through all the pre-commit hooks and pass or fail a PR based on the result. @amakarudze you can install this CI check by going to https://pre-commit.ci and choosing "sign in with github". Then select this repo to install the service.

I also had to add a `flake8` ignore, because it conflicted with `black`.

Then the final thing I did was to run the whole project through the `pre-commit` hooks as they are now. So the biggest commit here is python-upgrade and django-upgrade doing their job to get the project up to date.


ramonsaraiva:
I'd rather split this into two PRs, 1 running the existing hooks as they are (black, etc..) and then another one adding `django-upgrade` and `python-upgrade` pre-commit hooks, for this specific initial run. That way it is easier to distinguish between `black` and `django-upgrade`/`python-upgrade` automatic changes.

marksweb:
Ok I'll do it as 3 so it's nice & clear where changes come from.

PR #843, opened 2023-01-17
ci: Run pre-commit hooks on all files

marksweb:
Having introduced pre-commit hooks in #831 we need to have had those hooks run over the existing codebase so that it's up to the standard defined in the hooks.

This PR has a commit from me adding an ignored rule for `flake8` which conflict with `black` and then another which is completely automated by the hooks.

After this, I'll continue what I'd started in #842 because this will provide a "clean slate" from which to raise 2 more PRs for `django-upgrade` and then `python-upgrade`


ramonsaraiva (approved):
I'm happy I won't need to disable black when modifying files in this repo!

amakarudze (approved):

Issue #836, opened 2023-01-11
Add a sponsorship app to help track renewal dates for annual sponsorships


amakarudze:
Tracking sponsorship renewal is currently being done manually. Adding a sponsorship app will help the fundraising coordinator track sponsorship expiry dates and send automated emails.

PR #851, opened 2023-01-31
Global partners

amakarudze:
Changes in this PR
- Adds a `globalpartners` app for managing global partners and sending emails from the app
- Removes `Company` model from `jobboard/models.py`
- Alters field `company` of `Job` model to point to `GlobalPartner`


marksweb (approved):
This looks good.

My only feedback is just something I'd do differently - I'd store the sponsor level values as numbers rather than the strings. If any reporting were ever done on the data, numbers for how much people donate would be easier to query than the strings.

amakarudze:
Closes #836 

> This looks good.
> 
> My only feedback is just something I'd do differently - I'd store the sponsor level values as numbers rather than the strings. If any reporting were ever done on the data, numbers for how much people donate would be easier to query than the strings.

Thanks @marksweb. I have changed the sponsor levels to integers.

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+from .emails import (
+    send_promotional_material_email,
+    send_prospective_sponsor_email,
+    send_renewal_email,
+    send_thank_you_email,
+)
+
+PATREON_LEVELS = [
+    (100, 100),
```
marksweb, line 12:
```suggestion
    (100, "$100/month"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (250, 250),
```
marksweb, line 13:
```suggestion
    (250, "$250/month"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (500, 500),
```
marksweb, line 14:
```suggestion
    (500, "$500/month"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (1000, 1000),
```
marksweb, line 15:
```suggestion
    (1000, "$1,000/month"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+]
+
+ANNUAL_SPONSOR_LEVELS = [
+    (500, 500),
```
marksweb, line 19:
```suggestion
    (500, "Bronze ($500)"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (1000, 1000),
```
marksweb, line 20:
```suggestion
    (1000, "Silver ($1000)"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (2500, 2500),
```
marksweb, line 21:
```suggestion
    (2500, "Gold ($2,500)"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (5000, 5000),
```
marksweb, line 22:
```suggestion
    (5000, "Platinum ($5,000)"),
```

globalpartners/models.py
```diff
@@ -0,0 +1,68 @@
...
+    (10000, 10000),
```
marksweb, line 23:
```suggestion
    (10000, "Diamond ($10,000)"),
```

marksweb:
@amakarudze Sorry I should have given you an example rather than assume you'd know what I meant.

I'd make the choices;

```
[
    (100, "$100/month"),
    (250, "$250/month"),
    (500, "$500/month"),
    (1000, "$1,000/month"),
]
```

Then the database will store the integer, but where the user sees the choices they'll get the neat string to let them know the currency and the frequency of the payment.

I've added suggestions to the review to hopefully help out - you should just be able to accept each rather than actually making code changes.

I'd be good to also delete both migrations and generate the initial again so there's just 1.

marksweb (changes requested):

amakarudze:
Is there a reason for pre-commit fa

> @amakarudze Sorry I should have given you an example rather than assume you'd know what I meant.
> 
> I'd make the choices;
> 
> ```
> [
>     (100, "$100/month"),
>     (250, "$250/month"),
>     (500, "$500/month"),
>     (1000, "$1,000/month"),
> ]
> ```
> 
> Then the database will store the integer, but where the user sees the choices they'll get the neat string to let them know the currency and the frequency of the payment.
> 
> I've added suggestions to the review to hopefully help out - you should just be able to accept each rather than actually making code changes.
> 
> I'd be good to also delete both migrations and generate the initial again so there's just 1.

Yeah I thought about that and thought it would be me using it anyway and just used numbers all the way.

marksweb:
@amakarudze The pre-commit failure appears to be a problem with isort

I think an update of the pre-commit hooks is in order. I'll raise a PR.

PR #858, opened 2023-02-23
Add ssh deployment script

amakarudze:
Changes in this PR
- Add GitHub Action for running the deploy.sh script.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,27 @@
-- name: deploy to 🐍anywhere 
-  uses: appleboy/ssh-action@v0.1.7
-  with:
-    host: ${{ secrets.HOST }}
-    username: ${{ secrets.USERNAME }}
-    key: ${{ secrets.KEY }}
-    port: ${{ secrets.PORT }}
-    script: |
-      ./deploy.sh
+name: Deploy to 🐍anywhere
+
+on:
+  release:
+    types:
+      - published
+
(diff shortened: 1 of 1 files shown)


marksweb:
This should help improve processes 👍

marksweb (changes requested):
I think this will fail as the syntax here looks like it's just the step rather than the full file.

Take a look at an example here; https://github.com/marksweb/django-bleach/blob/main/.github/workflows/publish-to-live-pypi.yml

I think what you need is something like this;

```
name: Deploy to 🐍anywhere

on:
  release:
    types:
      - published

jobs:
  deploy:
    name: Deploy to 🐍anywhere 
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v3
    - name: Set up Python 3.9
      uses: actions/setup-python@v4
      with:
        python-version: 3.10

    - name: Run deploy script
      uses: appleboy/ssh-action@v0.1.7
      with:
        host: ${{ secrets.HOST }}
        username: ${{ secrets.USERNAME }}
        key: ${{ secrets.KEY }}
        port: ${{ secrets.PORT }}
        script: |
          ./deploy.sh
```

And with this, creating a release [here](https://github.com/DjangoGirls/djangogirls/releases) will trigger this workflow

marksweb (approved):

marksweb:
I _think_ this should be ok now 👍

PR #862, opened 2023-03-02
Use new version of the Mapbox JS library

### LINKED ISSUE #855: Events map broken
The map on https://djangogirls.org/en/events/map/ is now broken and only shows markers with no map. Inspecting the page indicates that the map no longer exists. It looks like the map has been deleted and needs to be created again so the page can work.

```
Request URL: https://a.tiles.mapbox.com/v4/olasitarska.m8nged0f.json?access_token=token&secure
Request Method: GET
Status Code: 410 Gone (from disk cache)
Remote Address: 52.85.219.178:443
Referrer Policy: same-origin
```

aniav:
Mapbox has deprecated their [Mapbox JS](https://docs.mapbox.com/mapbox.js/api/v3.3.1/) library. The map of events stopped working.

This PR:
- removes references to the old JS library and it's configs
- introduces the [mapbox GL JS](https://docs.mapbox.com/mapbox-gl-js/guides/) library
- adds a helper property returning X, Y coordinates required by the new library, instead of our standard lat, lng that worked with the previous version

The new map looks like this:
Screenshot 2023-03-02 at 22 07 56

Closes #855.


core/views.py
```diff
@@ -145,11 +139,7 @@ def year_2016_2017(request):
         request,
         "core/2016-2017.html",
         {
-            "events_2015": Event.objects.public().filter(date__lt="2016-01-01").order_by("date"),
```
aniav, line 148: This separate query wasn't used in the template at all. Must have been some kind of a leftover.

templates/core/2016-2017.html
```diff
@@ -78,7 +85,7 @@ <h2>{% trans "✨ Highlights ✨" %}</h2>
             <div id="growth" class="col-md-8 col-md-offset-2 col-sm-10 col-sm-offset-1 col-xs-10 col-xs-offset-1">
               <h2>{% trans "Growth 📈" %}</h2>
               <div id="map"></div>
-              <small>{% trans "Map of Django Girls event from July 2014 to July 2017" %}</small>
+              <small>{% trans "Map of Django Girls events from January 2016 to July 2017" %}</small>
```
aniav, line 88: This description was referring to different data than were loaded to the map, so I updated it to match what is pulled from the db.

templates/core/map.html
```diff
@@ -0,0 +1,91 @@
+{% load static %}
+
+<script>
+    mapboxgl.accessToken = 'pk.eyJ1IjoiZGphbmdvZ2lybHMiLCJhIjoiY2xjeWU1dm9iMG10cTNudGdrYTBkYm0ycSJ9.Ah4y76VVmBWqzVkw3kUzFQ'
```
amakarudze, line 4: Thanks for all your work on this @aniav. The only concern I have is having the public key publicly displayed. All our keys are now environment variables. It would be great to have this as an environment variable as well.

Otherwise, great work!

amakarudze (changes requested):
Great work on this. Just one minor detail on the public key in template instead of environment variable.

templates/core/map.html
aniav: So is this about the key being public or about having keys configurable per environment? Because this key is intended to be publiczny visible. There is no way to pass a key through a JS script in a hidden way. Mapbox suggests configuring the domains that can use the key to make it more secure.

If it’s about making the key configurable I'll change that 🙂

amakarudze: Yes, making it configurable would be great, thanks.

PR #869, opened 2023-03-07
Catch HttpError when creating/migrating emails

amakarudze:
When triaging event applications and creating emails for existing events, Google API often throws HttpError when the email exists and was not migrated properly. This PR catches the HttpError to prevent server errors as a result of these HttpErrors.


marksweb (approved):
LGTM 👍

PR #882, opened 2023-04-23
Fix logo fuzziness

dostogircse171:
This pull request fixes the issue with the logo looking fuzzy on retina screens. The logo has been updated to a high-resolution format and the dimensions have been adjusted to ensure that it looks clear on all screens.


static/source/css/global.styl
```diff
@@ -159,6 +159,9 @@ form
     .logo
         float: left
 
+        img
```
amakarudze, line 162: This class is not used on the logo in `templates/global/menu.html` where the CSS class for the logo is set as `logo`. Since you changed the size of the logo, this will have breaking changes on the navbar unless you set the width to the `logo` class in this file for navbar or edit the logo in `templates/global/menu.html` to use both classes `img` and `logo` or set the width/height of the new image to the dimensions of the deleted file.

amakarudze (changes requested):
Thanks for your work on this. Please see my comments.

PR #886, opened 2023-05-16
Add unique_together for name and twitter_handle for coach model

amakarudze:
Organizers keep adding coaches who already exist in the database and this is resulting in duplicate rows and hence we keep getting these `Coach.MultipleObjectsReturned: get() returned more than one Coach -- it returned 3!` errors due to data errors. This pull request seeks to add validation to new rows being added by making `name` and `twitter_handle` unique together and raise a validation error when such a row already exists,

Addresses #757 

Changes in this pull request:
- Set `DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"` in `settings.py` to stop ID autofield migrations when making changes to a model.
- Make `name` and `twitter_handle` unique together to avoid duplicate rows in coach model.
- Add tests for changes to coach model.

**NB**: Before merging this in, a cleanup exercise will be required to remove duplicate rows and keep the one with the lower id.


marksweb (approved):
Sorry for the delay!

PR #887, opened 2023-05-16
Catch ValueError due to approximate date with no day

### LINKED ISSUE #885: ValueError: day is out of range for month
Sentry Issue: [DJANGO-GIRLS-WEBSITE-5B](https://django-girls.sentry.io/issues/4131924376/?referrer=github_integration)

```
ValueError: day is out of range for month
(5 additional frame(s) were not displayed)
...
  File "organize/views.py", line 50, in done
    application = EventApplication.object.create(**data_dict)
  File "organize/models.py", line 53, in create
    if date(event_date.year, event_date.month, event_date.day) - date(
```

amakarudze:
If an existing organizer applies to organize an event while their old event application had an approximate date with month and year only, validation fails because the day which will be  will be out of range for that month. This PR fixes #885 

Changes in this PR
- Add `try...except` block to the event application validation and catch the `ValueError`.
- Add test for event application years apart.
- Add test for old event application with only an approximate date with `month` and `year` only.

### DIFF
--- a/organize/models.py
+++ b/organize/models.py
@@ -49,17 +49,29 @@ class EventApplicationManager(models.Manager):
         )
 
         if previous_event:
-            event_date = data_dict["date"]
-            if date(event_date.year, event_date.month, event_date.day) - date(
-                previous_event.date.year, previous_event.date.month, previous_event.date.day
-            ) < timedelta(days=180):
-                raise ValidationError(
-                    {
-                        "date": _(
-                            "Your workshops should be at least 6 months apart. " "Please read our Organizer Manual."
-                        )
-                    }
-                )
+            try:
+                event_date = data_dict["date"]
+                if date(event_date.year, event_date.month, event_date.day) - date(
+                    previous_event.date.year, previous_event.date.month, previous_event.date.day
+                ) < timedelta(days=180):
+                    raise ValidationError(
+                        {
+                            "date": _(
+                                "Your workshops should be at least 6 months apart. " "Please read our Organizer Manual."
+                            )
+                        }
+                    )
+            except ValueError:
+                if date(event_date.year, event_date.month, event_date.day) - date(
+                    previous_event.date.year, previous_event.date.month, 1
+                ) < timedelta(days=180):
+                    raise ValidationError(
+                        {
+                            "date": _(
+                                "Your workshops should be at least 6 months apart. " "Please read our Organizer Manual."
+                            )
+                        }
+                    )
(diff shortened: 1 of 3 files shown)


organize/models.py
```diff
@@ -49,17 +49,29 @@ def create(self, **data_dict):
...
-            ) < timedelta(days=180):
-                raise ValidationError(
-                    {
-                        "date": _(
-                            "Your workshops should be at least 6 months apart. " "Please read our Organizer Manual."
-                        )
-                    }
-                )
+            try:
+                event_date = data_dict["date"]
```
marksweb, line 53: I'd move this outside of the `try` because it's used by the `try` and `except`.

organize/models.py
```diff
@@ -49,17 +49,29 @@ def create(self, **data_dict):
...
+                    raise ValidationError(
+                        {
+                            "date": _(
+                                "Your workshops should be at least 6 months apart. " "Please read our Organizer Manual."
+                            )
+                        }
+                    )
+            except ValueError:
+                if date(event_date.year, event_date.month, event_date.day) - date(
+                    previous_event.date.year, previous_event.date.month, 1
```
marksweb, line 66: Is the `previous_event` day the only possible issue with the data that'll be provided here?

marksweb (commented):
In terms of the test data, this looks good. I've just got a couple of points.

And sorry this took so long. I just found this review request at the end of my notifications.

organize/models.py
amakarudze: Noted and done.

organize/models.py
amakarudze: Not sure but the error was resulting due to `previous_event` being an absolute date with just month and year. I guess maybe we could cater for both scenarios @marksweb ?

marksweb: Yeah, that gets a bit tricky though I bet. Maybe deploy this as is and we can keep an eye on errors should it need improvement.

organize/models.py
marksweb: @amakarudze I can't see any changes pushed. Will keep an eye out tomorrow 👍

organize/models.py
amakarudze: Noted.

organize/models.py
amakarudze: Sorry I forgot to push the commit. Just did @marksweb.

marksweb (approved):
LGTM 👍

Issue #888, opened 2023-05-16
Remove Google Groups mention and change instructions for Slack in new organizer email


amakarudze:
When a new event is deployed, organizers are told to join organizers' Google Group which no longer exists. They are also told that a Slack invite has been sent to them when at the moment we are unable to send automatic Slack invites. The email template for new events email should reflect this change.

Templates to be edited;
- [templates/emails/new_user.html](https://github.com/DjangoGirls/djangogirls/blob/main/templates/emails/new_user.html)
- [templates/emails/existing_user.html](https://github.com/DjangoGirls/djangogirls/blob/main/templates/emails/existing_user.html)
- [templates/emails/event_deployed.html](https://github.com/DjangoGirls/djangogirls/blob/main/templates/emails/organize/event_deployed.html)

amakarudze:
Yes, sure @anaschwendler. Please do.  Be sure to say *If you are not our Slack channel, please reach out to us and we will invite you and your team to Slack.* in place of *We have sent you a Slack invite.*.

anaschwendler:
I'm there, all good!
I'm on a work break, so I might be working on some issues. We can also speak if you want to send me some urgent topics!

See you, and have a lovely week!

amakarudze:
Yay! That's awesome @anaschwendler. Thanks, I will certainly reach out. See you and have a lovely week as well.

amakarudze:
Addressed by #894

PR #894, opened 2023-06-13
Remove Google Groups mention and change instructions for Slack in new organizer email

anaschwendler:
Solves #888

### DIFF
--- a/templates/emails/existing_user.html
+++ b/templates/emails/existing_user.html
@@ -31,11 +31,7 @@
 
 <p>
 {% blocktrans trimmed %}
-  You can contact hello@ if you need assistance but it’s usually faster to check
-  our <a href="https://faq-organizers.djangogirls.org/">FAQ</a> or to post questions
-  to the <a href="https://groups.google.com/forum/#!forum/django-girls-organizers">
-  Google Group of Django Girls Organizers</a> or <a href="https://djangogirls.slack.com/messages/general/">
-  Slack channel</a> - there are more people that can help you! :)
+  You can contact hello@ if you need assistance but it's usually faster to check <a href="https://djangogirls.slack.com/messages/general/">Slack channel</a> - there are more people that can help you! :)
 {% endblocktrans %}
 </p>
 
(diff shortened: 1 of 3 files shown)


templates/emails/existing_user.html
amakarudze, line 1: Hey @anaschwendler, so this part should read like: 

*You can contact hello@ if you need assistance but it's usually faster to check <a href="https://djangogirls.slack.com/messages/general/">Slack channel</a> - there are more people that can help you! :)*

We no longer have the Google Groups for Django Girls Organizers. I am sorry if my instructions wasn't so clear.

templates/emails/new_user.html
amakarudze, line 1: This part should read:

*You can contact hello@ if you need assistance but it's usually faster to check <a href="https://djangogirls.slack.com/messages/general/">Slack channel</a> - there are more people that can help you! :)*

templates/emails/organize/event_deployed.html
amakarudze, line 1: This part should read like this:

*Django Girls has a growing community of more than 2346 organizers.
If you are a new organizer and not already Django Girls Slack, please reach out to hello@djangogirls.org and we will invite you to join -- this is where you can chat with other organisers, ask for their help or advice. Use it wisely and plenty!*

.pre-commit-config.yaml
amakarudze, line 1: Can we put a hold on this? Our server doesn't support Python 3.11 yet and so our production version is Python 3.10 so it would not be ideal to have development version different from production version. So can you revert this change and have Python version as 3.10 instead?

amakarudze (changes requested):
Thanks @anaschwendler for your work on this. May you please change the Python version for pre-commit back to Python 3.10 which we are using in production? I have also given you the suggested text we would like to have in the email templates. Can you please update the text in the templates?

Otherwise great work on this. Thanks again!

amakarudze (approved):
LGTM! Thanks for your work on this @anaschwendler. Will merge this in tomorrow 🥇 .

Issue #736, opened 2022-02-11
Clean up settings.py and .environment by removing unused variables like jobs email, store discount and meetups email 


amakarudze:
We don't have a store open for now and no longer have jobs and meetups email accounts but they are still in settings. The variables need to be removed from `settings.py` and environment variables.

amakarudze:
Hello @anaschwendler, do you think this is something you can tackle? I can help with variables to be removed.

anaschwendler:
On it now! 🍀

PR #895, opened 2023-06-16
Remove unused variables in settings.py and .environment

anaschwendler:
Solves #736

### DIFF
--- a/.environment-example
+++ b/.environment-example
@@ -26,12 +26,6 @@ export TRELLO_API_KEY=""
 
 export SENTRY_DSN=""
 
-export JOBS_EMAIL_USER=""
-export JOBS_EMAIL_PASSWORD=""
-
-export MEETUPS_EMAIL_USER=""
-export MEETUPS_EMAIL_PASSWORD=""
-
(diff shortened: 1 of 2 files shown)


.environment-example
```diff
@@ -26,18 +26,9 @@ export TRELLO_API_KEY=""
...
-export JOBS_EMAIL_USER=""
-export JOBS_EMAIL_PASSWORD=""
-
-export MEETUPS_EMAIL_USER=""
-export MEETUPS_EMAIL_PASSWORD=""
-
 export GAPPS_PRIVATE_KEY_ID=""
 export GAPPS_PRIVATE_KEY=""
 
-export STRIPE_PUBLIC_KEY=""
```
amakarudze, line 38: STRIPE keys are still needed  for our donations form so you need to put them back.

amakarudze (changes requested):
Thanks for this @anaschwendler.

STRIPE keys are required for our donations form so we would need them there so anyone who ever takes over the repo knows which variables are required in production if they ever have to do an upgrade.

amakarudze (approved):
LGTM!

PR #899, opened 2023-07-06
Make sure application return only one objecte

anaschwendler:
This PR looks forward to solving #890 

To solve this issue that `Application` is returning two objects, I decided to filter them by `number` and then order them by `-created` (where - refers to the inverted list), selecting the latest application created,  following the Application Model: https://github.com/DjangoGirls/djangogirls/blob/main/applications/models.py#L116.


amakarudze (approved):
Thanks for this @anaschwendler!

PR #908, opened 2023-08-23
fixed the issue ##901

mdehsan873:


marksweb:
If you add "fixes #901" to your description, it'll link up to the issue and automatically close it if this is merged.

applications/views.py
```diff
@@ -36,20 +37,34 @@ def apply(request, page_url):
...
+            return render(
+                request,
+                "applications/apply.html",
+                {
+                    "event": event,
+                    "menu": menu,
+                    "form_obj": form_obj,
+                },
+            )
+        except IntegrityError:
```
marksweb, line 54: Although this fixes the issue, it assumes that any `IntegrityError` is caused by this issue. Now this may be true at the moment, but things may change. So we could merge this and catch this issue at this point.

There is some validation on the form that appears to be trying to check against this same constraint, and that is where I'd prefer the fix to go; https://github.com/DjangoGirls/djangogirls/blob/main/applications/forms.py#L40

PR #909, opened 2023-08-23
Resolved issue #859. Fixed attribute error.

Nandika-A:
The error was caused at /events/eventname/applications due to selecting more than one way of ordering. 
It is resolved by giving a default value to getattr(), obtained from the function get_default(). This function slices the string to get the first order selected.


applications/services.py
```diff
@@ -3,6 +3,10 @@
 from applications.models import Application, Event, Score
 from core.models import User
 
+def get_default(order):
```
marksweb, line 6: Could you please add a docstring that explans what this is for & what it is doing?

marksweb (changes requested):

applications/services.py
Nandika-A: Ya sure!

marksweb (approved):
This looks straightforward enough. Would be ideal to have some testing with it though.

Nandika-A:
Hello @amakarudze! I have resolved the issues with flake8. Could you please check it now?

requirements.txt
```diff
@@ -1,331 +1,254 @@
-#
-# This file is autogenerated by pip-compile with Python 3.10
-# by the following command:
-#
-#    pip-compile --output-file=requirements.txt --resolver=backtracking requirements.in
-#
+absl-py==1.4.0
```
marksweb, line 1: It looks like these changes having been done using `pip-tools` is that correct?

`pip-compile` should be used to generate, and update the project requirements.

marksweb (changes requested):

requirements.txt
Nandika-A: Yes, I used pip-tools to update the virtual environment according to requirements as per the errors shown by flake8. Should I revert the changes in requirements.txt?

Nandika-A: I have restored requirements.txt. Could you please check it now?

marksweb: Thanks. It's not a problem to update the packages. It's just that it should have the comment about it being a generated file and then the packages that use each defined package underneath them.

marksweb (approved):

amakarudze:
Thanks @Nandika-A. Now merging this in.

PR #917, opened 2023-10-28
Add event funding information

amakarudze:
Changes in this PR
- Add event funding template, view, url and tests

### DIFF
--- a/templates/organize/event_funding.html
+++ b/templates/organize/event_funding.html
@@ -37,7 +37,7 @@
         </p>
 
         <p>To apply for DSF funding, please fill in 
-            <a href="https://docs.google.com/forms/d/1KBQKnwY0uFBB0R7HOJIA3VAr9PsUywnp-9yX9El9-Wk/viewform?edit_requested=false">this form</a>.
+            <a href="https://docs.google.com/forms/d/1KBQKnwY0uFBB0R7HOJIA3VAr9PsUywnp-9yX9El9-Wk/viewform">this form</a>.
         </p>
     </div>
 </div>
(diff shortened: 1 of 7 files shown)


templates/organize/event_funding.html
```diff
@@ -0,0 +1,44 @@
+{% extends "global/base.html" %}
+
+{% block content %}
+<div class="container">
+    <h2>Apply for Sponsorship from our Global Partners</h2>
```
marksweb, line 5: I suspect this should have `{% translate` tags in it so that we can provide it in the available languages?

tests/organize/test_views.py
```diff
@@ -106,3 +106,8 @@ def test_organize_form_wizard_workshops_too_close(client, previous_organizer_in_
                 response["form.errors"] == "Your workshops should be at least 6 months apart. "
                 "Please read our Organizer Manual."
             )
+
+
+def test_event_funding(client):
+    response = client.get(reverse("organize:event_funding"))
```
marksweb, line 112: I think it'd be worth verifying the template used here as well.

```
def test_event_funding(client):
    response = client.get(reverse("organize:event_funding"))
    assert response.templates[0].name == "organize/event_funding.html"
    assert response.status_code == 200
```

marksweb (changes requested):

amakarudze:
Thanks, @marksweb, I have been busy with DjangoCon Africa, so I can now look at this. 

Flake 8 is passing locally but failing for all new pull requests and what it's suggesting seems pretty weird to me. Should we consider changing this action @marksweb?

marksweb:
@amakarudze yeah I see. With the slack message in the management utils it might be complaining about the slack emojis.

I'd have to have a look from my laptop to be sure. But I can look at ignoring things that are actually valid or changing for the ruff linter which is what use these days.

marksweb:
Ok I got ruff in quicker than expected. #923

PR #926, opened 2023-12-31
Fix #921: Truncate latlng to 30 chars

### LINKED ISSUE #921: DataError: value too long for type character varying(30)
Sentry Issue: [DJANGO-GIRLS-WEBSITE-60](https://django-girls.sentry.io/issues/4635355583/?referrer=github_integration)

```
StringDataRightTruncation: value too long for type character varying(30)

  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)

DataError: value too long for type character varying(30)

(20 additional frame(s) were not displayed)
...
  File "organize/views.py", line 50, in done
    application = EventApplication.object.create(**data_dict)
  File "organize/models.py", line 75, in create
    return super().create(**data_dict)
  File "organize/models.py", line 146, in save
    super().save(*args, **kwargs)
```

dostogircse171:
This pull request fixes #921  the DataError encountered when saving `latlng` values longer than 30 characters. 
It modifies the `save` method of the `EventApplication` model to truncate the `latlng` value to 30 characters, ensuring data integrity without altering the database schema.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -15,7 +15,9 @@ def get_coordinates_for_city(city, country):
 
     try:
         data = req.json()[0]
-        return f'{data["lat"]}, {data["lon"]}'
+        formatted_lat = "{:.6f}".format(float(data["lat"]))
+        formatted_lon = "{:.6f}".format(float(data["lon"]))
+        return f'{formatted_lat}, {formatted_lon}'
     except (IndexError, KeyError):
         return None
 
--- a/core/utils.py
+++ b/core/utils.py
@@ -15,9 +15,9 @@ def get_coordinates_for_city(city, country):
 
     try:
         data = req.json()[0]
-        formatted_lat = "{:.6f}".format(float(data["lat"]))
-        formatted_lon = "{:.6f}".format(float(data["lon"]))
-        return f'{formatted_lat}, {formatted_lon}'
+        formatted_lat = "{:.7f}".format(float(data["lat"]))
+        formatted_lon = "{:.7f}".format(float(data["lon"]))
+        return f"{formatted_lat}, {formatted_lon}"
     except (IndexError, KeyError):
         return None
 
--- a/organize/models.py
+++ b/organize/models.py
@@ -129,8 +129,7 @@ class EventApplication(models.Model):
 
     def save(self, *args, **kwargs):
         if not self.latlng:
-            latlng_coord = get_coordinates_for_city(self.city, self.get_country_display())
-            self.latlng = latlng_coord[:30]  # Truncate to 30 characters to match the database field limit
+            self.latlng = get_coordinates_for_city(self.city, self.get_country_display())
 
         super().save(*args, **kwargs)
 
--- a/organize/models.py
+++ b/organize/models.py
@@ -129,7 +129,8 @@ class EventApplication(models.Model):
 
     def save(self, *args, **kwargs):
         if not self.latlng:
-            self.latlng = get_coordinates_for_city(self.city, self.get_country_display())
+            latlng_coord = get_coordinates_for_city(self.city, self.get_country_display())
+            self.latlng = latlng_coord[:30]  # Truncate to 30 characters to match the database field limit
 
         super().save(*args, **kwargs)
 
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -24,7 +24,9 @@ def test_get_coordinates_for_city(mock_get):
             "q": "London, UK",
         },
     )
-    assert result == "1.23, 4.56"
+    expected_lat = "{:.7f}".format(float("1.23"))
+    expected_lon = "{:.7f}".format(float("4.56"))
+    assert result == f"{expected_lat}, {expected_lon}"
 
 
 @mock.patch("requests.get")


organize/models.py
```diff
@@ -129,7 +129,8 @@ def __str__(self):
 
     def save(self, *args, **kwargs):
         if not self.latlng:
-            self.latlng = get_coordinates_for_city(self.city, self.get_country_display())
+            latlng_coord = get_coordinates_for_city(self.city, self.get_country_display())
+            self.latlng = latlng_coord[:30]  # Truncate to 30 characters to match the database field limit
```
amakarudze, line 133: If we get values like this `(-74.80111590000001, -74.80111590000001) of which each value is 19 characters, we would truncate this and have the first value remain 19 characters while the second value becomes -74.80111 is 9 characters long. It would be better to format the values in the `get_coordinates_for_city()` and limit the number of decimals places for each so that they are of equal length, making consideration for the possible `-` in the values. The function is here  https://github.com/DjangoGirls/djangogirls/blob/main/core/utils.py#L18 in `core/utils.py`.

Thanks for your work on this.

amakarudze (changes requested):

organize/models.py
dostogircse171: Thank you very much for the review and letting me know @amakarudze 
I Got it. Let me work on this and I will get back soon.

dostogircse171: I have made those changes. Please let me know if any changes are required.

amakarudze (approved):
Thanks for your work on this @dostogircse171 !

core/utils.py
```diff
@@ -15,7 +15,9 @@ def get_coordinates_for_city(city, country):
 
     try:
         data = req.json()[0]
-        return f'{data["lat"]}, {data["lon"]}'
+        formatted_lat = "{:.6f}".format(float(data["lat"]))
+        formatted_lon = "{:.6f}".format(float(data["lon"]))
+        return f'{formatted_lat}, {formatted_lon}'
```
amakarudze, line 20: The ruff build is failing because of the single quotes here. May you please address this so we can merge this pull request. Sorry for taking so much time to review again.

organize/models.py
amakarudze: Thanks, sorry for not getting back to you earlier on. Just have one more comment below.

core/utils.py
dostogircse171: Hi @amakarudze ,
Thank you for your review. No problem let me fix this and I will update the code soon.

organize/models.py
amakarudze: One more change, can you truncate to 7 decimal places, tests are failing because the values in the tests are set to 7 decimal places while your changes truncate to 6 decimal places. Truncating to 7 decimal places should fix this. Also use double quotes for the `f` string to address ruff issues.

amakarudze (changes requested):
Please address the 7 decimal places and single quotes issues. Thanks!

core/utils.py
amakarudze: You're welcome. Thanks for your work on this.

organize/models.py
dostogircse171: Yes sure. Will do. Thank you

dostogircse171:
> Please address the 7 decimal places and single quotes issues. Thanks!

This is done. Let me know if anything else is required here. 
Thank you

amakarudze (approved):
LGTM!

dostogircse171:
@amakarudze I think we also need to adjust test_utils.py to match new changes.

amakarudze:
> @amakarudze I think we also need to adjust test_utils.py to match new changes.

Yeah, true I just noticed it has 8 decimal places, so removing 1 each should work. Sorry about that.

dostogircse171:
> > @amakarudze I think we also need to adjust test_utils.py to match new changes.
> 
> Yeah @dostogircse171 , true I just noticed it has 8 decimal places, so removing 1 each should work. Sorry about that.

@amakarudze I can see inside test_utils.py file same function 
`test_day_before_deadline()` has been used 3 times with the same values. is this a mistake or it has some purpose?

amakarudze:
> > > @amakarudze I think we also need to adjust test_utils.py to match new changes.
> > 
> > 
> > Yeah @dostogircse171 , true I just noticed it has 8 decimal places, so removing 1 each should work. Sorry about that.
> 
> @amakarudze I can see inside test_utils.py file same function `test_day_before_deadline()` has been used 3 times with the same values. is this a mistake or it has some purpose?

@dostogircse171 i have no idea, will have to look into it first before making any changes. Thanks for your work on this, I have merged the PR and some of my issues will go away because of this change!

dostogircse171:
> > > > @amakarudze I think we also need to adjust test_utils.py to match new changes.
> > > 
> > > 
> > > Yeah @dostogircse171 , true I just noticed it has 8 decimal places, so removing 1 each should work. Sorry about that.
> > 
> > 
> > @amakarudze I can see inside test_utils.py file same function `test_day_before_deadline()` has been used 3 times with the same values. is this a mistake or it has some purpose?
> 
> @dostogircse171 i have no idea, will have to look into it first before making any changes. Thanks for your work on this, I have merged the PR and some of my issues will go away because of this change!

@amakarudze Thank you very much. It's a pleasure to work on this task. Is there any other task you can recommend me to pick next?

PR #940, opened 2024-02-01
ci: Cleanup linting & switch to ruff formatter

marksweb:
This improves the efficiency of the current ruff github action and switches black for the ruff formatter which is effectively the same in how it formats code, but faster and with a few configuration options.

I've also removed autoflake as not to conflict with the ruff linter.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Ruff
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/ruff.yml
```diff
@@ -15,16 +15,11 @@ jobs:
...
-          python-version: "3.10"
-          cache: 'pip'
-      - run: |
-          python -m pip install --upgrade pip
-          pip install ruff
-      - name: Run Ruff
-        run: ruff .
+    - uses: actions/checkout@v4
+
+    - run: python -Im pip install --user ruff
```
amakarudze, line 22: Is this a typo `python -Im pip install --user ruff` should it be `python -m pip install --user ruff`?

amakarudze (changes requested):
I think you have a typo in one of the files so I left a comment there.

.github/workflows/ruff.yml
marksweb: No that's intentional. It's python's isolated mode flag

https://docs.python.org/3/using/cmdline.html#cmdoption-I

amakarudze: Noted, with thanks. Let me approve this.

amakarudze (approved):

PR #943, opened 2024-02-05
Add Google Tag Manager and cookie consent

amakarudze:
Changes in this PR:
- Add a tag for Google Analytics to enable tracking website visits.
- Add cookie consent for users.

The cookie consent dialogue looks as below:
Screenshot 2024-02-05 at 17 13 32

### DIFF
--- a/templates/event/base.html
+++ b/templates/event/base.html
@@ -3,6 +3,14 @@
 <!DOCTYPE HTML>
 <html>
 <head>
+  <!-- Google tag (gtag.js) -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-JGJW1VEY0D"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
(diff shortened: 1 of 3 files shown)


templates/event/base.html
```diff
@@ -3,6 +3,14 @@
 <!DOCTYPE HTML>
 <html>
 <head>
+  <!-- Google tag (gtag.js) -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-JGJW1VEY0D"></script>
```
marksweb, line 7: This gets loaded regardless of the cookie consent choice, so what functionality is listening out for that consent?

amakarudze: Thanks, let me add that functionality.

amakarudze: @marksweb, I followed the guidelines here https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out to disable the google tag on cookie reject.

marksweb (approved):
Ok, that looks good 👍

PR #950, opened 2024-02-18
Handle 404 errors in fetch_stories

### LINKED ISSUE #949: HTTPError 404: Not Found Exception When Fetching Images in fetch_stories Command
When running the `python manage.py fetch_stories` command to import blog posts from the Django Girls RSS feed, it encounters an `HTTPError 404: Not Found` exception for some posts when attempting to download associated images. This error halts the command, preventing further posts from being processed.

For example: [Django Girls CDO invades Bukidnon](https://blog.djangogirls.org/post/187602084318/django-girls-cdo-invades-bukidnon) blog post has a broken image which is failing while fetching the data due to this when I tried. 

**Reproduce the error**: This issue can be produced by deleting all the stories from Django admin and then running the command `python manage.py fetch_stories` as per the below screenshot:

Screenshot 2024-02-18 at 3 17 21 am

To solve this, We can Implement exception handling for HTTPError around the image download logic to allow the command to log the issue and continue with the next post.

I have already worked on this issue and will do a PR for review.

dostogircse171:
This PR addresses the issue encountered when fetching blog posts with images that result in a 404 error. By implementing error handling for these cases, the fetch_stories command can now gracefully skip over images that are not found, ensuring that the process continues to fetch and store all available blog posts without interruption and Image (If 404).

This approach ensures that individual failures due to missing images do not halt the overall execution, improving the reliability and robustness of our blog post fetching mechanism.

Screenshot 2024-02-18 at 3 26 52 am

Fixes #949

### DIFF
--- a/story/management/commands/fetch_stories.py
+++ b/story/management/commands/fetch_stories.py
@@ -5,6 +5,7 @@ from django.core.files import File
 from django.core.files.temp import NamedTemporaryFile
 from django.core.management.base import BaseCommand
 from pyquery import PyQuery as pq  # noqa: N813
+from urllib.error import URLError, HTTPError
 
 from story.models import Story
 
@@ -44,7 +45,7 @@ class Command(BaseCommand):
                             img.write(urlopen(image_url).read())
                             img.flush()
                             story.image.save(image_url.split("/")[-1], File(img))
-                    except Exception as e:
+                    except (HTTPError, URLError, Exception) as e:
                         print(f"Failed to fetch image from {image_url}: {e}")
 
                 story.save()
--- a/story/management/commands/fetch_stories.py
+++ b/story/management/commands/fetch_stories.py
@@ -1,4 +1,4 @@
-from urllib.error import HTTPError, URLError
+from urllib.error import URLError, HTTPError
 from xml.etree import ElementTree
 
 import requests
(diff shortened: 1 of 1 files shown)


story/management/commands/fetch_stories.py
```diff
@@ -39,10 +39,13 @@ def handle(self, *args, **options):
...
-                    img = NamedTemporaryFile(delete=True)
-                    img.write(urlopen(image_url).read())
-                    img.flush()
-                    story.image.save(image_url.split("/")[-1], File(img))
+                    try:
+                        with NamedTemporaryFile(delete=True) as img:
+                            img.write(urlopen(image_url).read())
+                            img.flush()
+                            story.image.save(image_url.split("/")[-1], File(img))
+                    except Exception as e:
```
marksweb, line 47: This will catch all exceptions, which is better than catching none, but by catching all, the printed message may not actually be true.

I'd rather see specific exceptions caught with appropriate messages for them.

marksweb (changes requested):

story/management/commands/fetch_stories.py
dostogircse171: @marksweb Thank you very much for your review. You are right.

Should I primarily focus on handling exceptions related to the image not being found, as that is the main reason for failures here? 
Would you suggest catching specific exceptions for scenarios like a missing image, or are there other potential errors I should consider as well?

marksweb: I think catching the obvious errors is a good idea, catching the HTTP error is the easy one. Then if others occur we can catch those.

marksweb (approved):

PR #967, opened 2024-04-05
Update django.po

anitahammer:


marksweb:
I don't think this is a beneficial or worthwhile change. Also, we don't directly modify these files, we manage translations through poeditor here:

https://poeditor.com/join/project/n5I3liMVyj

PR #982, opened 2024-04-22
feat: Allow translation management from docker

marksweb:
This gets docker to run `makemessages` and `compilemessages` which should help to maintain a consistent way to manage the message catalog and keep it up to date as people are working.

I thought about having the two commands in `entrypoint.sh` commented out and then uncommented when people want to run them, but that will lead to them never being ran.

I'd rather people had these files generatead and had to think about if they need to commit changes, than changes not getting created and the codebase falling behind POeditor or conflicts being created again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,25 +2,34 @@ FROM python:3.10.9-bullseye
 
 RUN apt-get update \
     && apt-get upgrade -yq \
+    && apt-get -y install locales \
+    && apt-get -y install gettext \
+    && apt-get -y install poedit \
     && apt-get install -yq --no-install-recommends \
         nodejs \
-        npm
+        npm \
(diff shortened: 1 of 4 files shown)


vanessa:
Oddly enough, I am currently unable to run the app now because of a failure in the `compilemessages` step.

```
djangogirls-app  | Execution of msgfmt failed: /var/www/app/locale/de/LC_MESSAGES/django.po:1066: 'msgid' and 'msgstr[1]' entries do not both begin with '\n'
djangogirls-app  | msgfmt: found 1 fatal error
```

Am I missing anything?

marksweb:
@vanessa ah this must have happened in a translation file since I put this together.

It's an easy fix though. Go to the referenced file and line number and you'll see an entry for a string. The source string likely starts with a `\n` new line character. The translated string needs to match that.

Issue #430, opened 2017-05-06
Second click in RSVP confirmation shows an error


jualvarez:
If a participant clicks twice on the RSVP confirmation link, the second time it throws an error.
Also related, once the RSVP is confirmed it cannot be changed to cancelled, it throws the same error.

The error says:
"Something went wrong with your RSVP link. Please contact us at <event email> with your name."

Participants get really anxious although the confirmation was received :)

das-g:
We (Django Girls Zürich) just got contacted by an applicant saying she got the message
> Something went wrong with your RSVP link. Please contact us at zurich@djangogirls.org with your name.

when clicking the RSVP link. (She didn't say anything about clicking it twice, but it's of course possible that that's what she did. Is there any way we as organizers can find out? It looks like her RSVP status was updated alright.)

UX-wise probably related to #121.

kewtree1408:
The current logic is as follows: when an applicant is accepted, their status becomes `waiting_rsvp`. Then, when the applicant clicks the RSVP link, this check occurs:
https://github.com/DjangoGirls/djangogirls/blob/main/applications/views.py#L302-L308.

Currently, we don’t display the existing `rsvp_status` stored in the database. I think we can enhance this by displaying the current RSVP status clearly on that page.
Something like: 

```python
        if application.rsvp_status == Application.RSVP_YES:
            message = _("You have already confirmed your attendance.")
        elif application.rsvp_status == Application.RSVP_NO:
            message = _(
                "You have already declined the invitation. ",
                "If you want to change your RSVP, ",
                "please contact us at %(email)s with your name.") % {"email": event.email}

        messages.error(
            request,
            message,
        )
```
 If you agree, I can proceed with implementing this improvement.

kewtree1408:
Although the PR has been merged and tested by myself and a few others, we’ve received reports that some RSVP links are not working. My current assumption is that in certain cases, the links might be broken, but without access to the database or logs, it’s difficult to investigate further.

If anyone has the opportunity to test the full RSVP flow again, that would be extremely helpful. Until then, we (Django Girls Amsterdam) have decided not to use this feature.

PR #1053, opened 2025-03-09
Second click in RSVP confirmation shows better messages

kewtree1408:
Fix this issue: https://github.com/DjangoGirls/djangogirls/issues/430


kewtree1408:
Could you advise on the best way to proceed with the .po files?

Locally, I ran the following commands:
```
python manage.py makemessages -l nl --ignore 'venv/*' 
python manage.py compilemessages -l nl 
```
All tests passed successfully, and no test changes were needed since these updates only affect messages. The redirect logic remains unchanged and continues to work as expected.

Let me know if you have any suggestions or if anything else needs to be done.

fltfx:
Looks good to me.

marksweb:
> Could you advise on the best way to proceed with the .po files?

I've put message creation into the docker setup. In `rootfs/entrypoint.sh` there is `#./manage.py makemessages --all`

Typically I add this command back in to generate new `po` files. We've then got a translations branch and an external service which handles the translations themselves.

marksweb:
If you could please rebase this & resolve the conflict, all checks should pass as I've just fixed them.

kewtree1408:
Hey @marksweb ,
I’ve just rebased the branch, and all the new checks have passed.

Regarding the .po files:
I noticed a commented-out command in the entry point, but there doesn’t seem to be any .po-related action in the CI/CD pipeline.

Would you like me to remove the `nl` files from this PR and regenerate the `.po` files for all languages in a separate branch?
Or would you prefer to keep things as they are?

marksweb:
@kewtree1408 Yeah there isn't any CI for translations yet, though I do like the sound of something helpful like that.

It may be easiest to remove the translations from this change, then generate them on the tranlstaions branch as I usually do.

kewtree1408:
Hey @marksweb,  I’ve removed the translations and formatting from the previous version, so it’s now cleaner. Do you have any other comments?

marksweb (approved):
@kewtree1408 No, this looks great now - thank you!

PR #1060, opened 2025-03-27
Remove hard-coded email for hello

amakarudze:
Remove the hard-coded value for hello@djangogirls.org to prevent emails from being sent to hello@djangogirls.org when people are playing around with the website locally.


contact/models.py
```diff
@@ -1,9 +1,12 @@
 from smtplib import SMTPException
 
+from django.conf import settings
 from django.core.mail import EmailMessage
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
+support_email = settings.DEFAULT_FROM_EMAIL
```
marksweb, line 8: Why create the variable at module level and not just refer to `settings.DEFAULT_FROM_EMAIL` in the code?

Also, module level variables should be uppercase.

amakarudze: My bad. I need to revisit the PEP8 guidelines. I removed the variables. I'm sorry I took too long to fix this. I got busy with school.

marksweb (approved):

PR #1104, opened 2026-03-25
fix: add distinct() to CoachAdmin queryset to prevent MultipleObjects…

### LINKED ISSUE #1075: Coach.MultipleObjectsReturned: get() returned more than one Coach -- it returned 7!
Sentry Issue: [DJANGO-GIRLS-WEBSITE-82](https://django-girls.sentry.io/issues/6656350128/?referrer=github_integration)

```
Coach.MultipleObjectsReturned: get() returned more than one Coach -- it returned 7!
(8 additional frame(s) were not displayed)
...
  File "django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/contrib/admin/options.py", line 1540, in changeform_view
    return self._changeform_view(request, object_id, form_url, extra_context)
  File "django/contrib/admin/options.py", line 1561, in _changeform_view
    obj = self.get_object(request, unquote(object_id), to_field)
  File "django/contrib/admin/options.py", line 763, in get_object
    return queryset.get(**{field.name: object_id})
  File "django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(
```

dostogircse171:
`CoachAdmin.get_queryset` filters coaches via `eventpagecontent__event__team`, which is a many-to-many join. When a coach is linked to multiple `EventPageContent` blocks, this join produces duplicate rows in the queryset. Django admin then calls `queryset.get(pk=object_id)` on that queryset when opening a coach record, which raises `MultipleObjectsReturned` because it finds the same coach multiple times.

Fixed by adding `.distinct()` to the queryset filter, so each coach appears only once regardless of how many `EventPageContent` blocks they are linked to. This is consistent with how `SponsorAdmin.get_queryset` already handles the same pattern.

Fixes #1075

### DIFF
--- a/coach/admin.py
+++ b/coach/admin.py
@@ -19,7 +19,7 @@ class CoachAdmin(admin.ModelAdmin):
         qs = super().get_queryset(request)
         if request.user.is_superuser:
             return qs
-        return qs.filter(eventpagecontent__event__team=request.user)
+        return qs.filter(eventpagecontent__event__team=request.user).distinct()
 
     def get_form(self, request, obj=None, **kwargs):
         form = super().get_form(request, obj, **kwargs)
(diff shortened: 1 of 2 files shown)


marksweb:
@dostogircse171 Thanks for this. I've updated some packages in `main` to resolve issues running the tests.

Could you please rebase your branch to allow the test suite to run?

PR #1120, opened 2026-08-18
Fix Docker static files issues

amakarudze:
This pull request fixes static file issues when using the Docker setup. Changes in this pull request include:

- Rename `.environment_example` to `.env-example`.
- Add `NODE_ENV` environment variable to `.env-example`.
- Add `env_file` and `node_modules` volume to `docker-compose.yaml`.
- Fix Dockerfile to use only Node version 24 and fix Python installation issues.
- Add `gulp watch` command to `entrypoint.sh` for development environment.
- Add `dotenv` to `gulpfile.js` and `package.json`.
- Remove the exposed Mapbox key from `.env-example`.
- Update `README.md` Docker setup section.


.env-example
marksweb, line 9: I think this needs to be removed. The compose file sets the database environment variables. So for this to work you'd have to duplicate that data.

marksweb (commented):
I'm getting a test failure that I'm assuming would be unrelated to these changes so I'm not sure why.

```
FAILED tests/applications/test_services.py::test_it_sorts_applications_by_score_excluding_hidden - assert [(False, 4.5)... (False, 1.5)] == [(True, 3.667... (False, 1.5)]
```

amakarudze:
> I'm getting a test failure that I'm assuming would be unrelated to these changes so I'm not sure why.
> 
> ```
> FAILED tests/applications/test_services.py::test_it_sorts_applications_by_score_excluding_hidden - assert [(False, 4.5)... (False, 1.5)] == [(True, 3.667... (False, 1.5)]
> ```

The test was flacky for me as well so I had to fix it. I have had some messed up git history so I will probably close this one and create a new one with less commits.

.env-example
amakarudze: docker-compose file only has database settings. I added a line to it to use .env so that the user can set DEBUG=TRUE to be able to generate local static files using gulp. Without it, I didn't have static files when using Docker. Since the .env-example is used even by users using a virtual environment, I will leave the file as is and maybe remove database settings from docker-compose to avoid repetition.

